PR #12, opened 2025-07-09
Add Automated Release Workflow

vs4vijay:
Just create a tag with pattern v0.0.0 and it would create a release with apk

Example:

image


callebtc:
awesome very useful thank you

Issue #17, opened 2025-07-09
Return Button Cut-off in ChannelHeader and PrivateChatHeader


deltaLuki:
## Description:
The return button displayed in both the `ChannelHeader` and `PrivateChatHeader` components is not fully visible. Part of the button's content is being cut off, making it difficult for users to clearly see or interact with it.

<img width="508" height="49" alt="Image" src="https://github.com/user-attachments/assets/66e1f739-1845-4bfb-8de4-c148df4fc2bf" />

## Proposed Solution:
This issue might be resolved by ensuring that an `Icon` and a `Text` element are wrapped within a `Row` component. 

<img width="508" height="49" alt="Image" src="https://github.com/user-attachments/assets/8b3b320d-d853-4176-abdb-b23f07dd1a60" />

callebtc:
fixed in #75

PR #18, opened 2025-07-09
Improve Header Return Button Layout in PrivateChatHeader and ChannelHeader

deltaLuki:
## Problem 
The return button displayed in both the ChannelHeader and PrivateChatHeader components is not fully visible. Part of the button's content is being cut off, making it difficult for users to clearly see or interact with it.

Issue: #17 

## Solution
Refactored the back button into a reusable composable HeaderReturnButton, which:
	•	Uses a proper Row layout for consistent alignment of Icon and Text
	•	Uses Icons.Default.ArrowBack


callebtc:
appreciate this very much and sorry for not noticing but I've fixed that in the last PR 🫡  would love more contributions like this!

callebtc:
I just saw it's still broken in the channels

PR #26, opened 2025-07-10
Use tag/build number as versionName/versionCode.

thesebas:
Example release https://github.com/thesebas/bitchat-android/releases/tag/0.3.0-1.
Also sets correct versionName and versionCode in compiled apk.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,9 +38,16 @@ jobs:
       - name: Grant execute permission for Gradlew
         run: chmod +x ./gradlew
 
+      - name: Set versionName from release tag
+        run: echo "VERSION_NAME=${GITHUB_REF##*/}" >> $GITHUB_ENV
+
       - name: Build APK
-        run: ./gradlew assembleRelease --no-daemon --stacktrace
-      
+        run: ./gradlew assembleRelease \
+          --no-daemon \
+          --stacktrace \
+          -PversionName=${{ env.VERSION_NAME }} \
+          -PversionCode=${{ github.run_id }}
+
       - name: List APK files
         run: |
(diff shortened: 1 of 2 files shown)


callebtc:
I just merged #12 but your PR seems nicer because of the versioncode and versionname... could we combine both? would really like this in

thesebas:
combined 🪄

callebtc:
thank you. just to be clear how this works, I create a tag with description etc and the pipeline creates a release with the same title, desciption, and the apk file?

thesebas:
The workflow comes from #12, and yeah, thats what I assume, tag creation (I would argue if it shuold be x.y.x or vx.y.z ;) ) triggers this workflow, and this workflow builds the app and creates the release with autogenerated description. My change only adds tag as versionName (and build number as versionNumber) to the already existing build process implemented in #12 by @vs4vijay.

thesebas:
rebased and resolved conflicts

vs4vijay:
@thesebas I’m good with the pattern x.y.x. Let’s go ahead and implement this in your PR and resolve any conflicts.

@callebtc You're correct, when you want to make a release, just create a tag manually from the GitHub UI using the specified pattern (vx.y.x or x.y.x). This will trigger the creation of a new release with the latest codebase and build the APK along with its description. There is some commented-out code related to signing the APK; if you want to enable signing, you’ll need to add the necessary secret, I’ve left a placeholder for that.

Let me know if you need any help on this front.

Chessing234 (commented):
github.run_id does not fit in versionCode (int32). toIntOrNull will miss and you ship versionCode 1 on every release. derive a small monotonic int from the tag, not the run id.

Issue #37, opened 2025-07-10
/pass command handling


alpermelkeli:
### Checklist

- [x] I have used the search function for [**OPEN**](https://github.com/callebtc/bitchat-android/issues) issues to see if someone else has already submitted the same feature request.
- [x] I have **also** used the search function for [**CLOSED**](https://github.com/callebtc/bitchat-android/issues?q=is%3Aissue+is%3Aclosed) issues to see if the feature was already implemented and is just waiting to be released, or if the feature was rejected.
- [x] I will describe the problem with as much detail as possible.
- [x] This request contains only one single feature, **not** a list of multiple (related) features.

### App version

Main Branch

### Feature or improvement you want

/pass command is not handled at code.
I implement something for this but there is issue that I cannot announce this room has password to other users. I want to discuss the business logic here.

BlasterStock:
I think that users should not be notified that a password has been set, but the fact that there is a password should be. That is, each new user, when using /j, will have to receive information that a password is needed to log in to the group

alpermelkeli:
yes it's make sense. How can we do that? Now user can join all channels without password.

alpermelkeli:
I have pr for /pass command handling so maybe I can add this feature to this pr.  #39

BlasterStock:
> yes it's make sense. How can we do that? Now user can join all channels without password.

A tricky question. If none of the group members are online, another person can create a group with the same name. What happens if there are two groups with identical names? Although group information can be shared by members with others who use `/j` to join the group (for example, they can communicate whether the group has a password), I think that if there are two groups with the same name but different passwords, the groups might break or overlap with each other

PR #39, opened 2025-07-10
implement channel password management at local and /pass command hand…

alpermelkeli:
Description
This PR implements support for the /pass command handling.

Currently, when a user tries to use the /pass command, it is not recognized or handled in the business logic. This feature allows a user to set or announce a password for the current channel.

The logic is designed to parse and validate the /pass command and then store the channel with it's password locally -> to announce this "channel has password from now" to other connected users we need to implement a logic like IOS project. (The IOS project has announce method inside mesh service class)

Closes: #37

How it works
Parses messages starting with /pass
Extracts the password string
Applies necessary validation (e.g., minimum length)
Triggers the appropriate logic to store the password but not for share this event other connected users

Checklist

- [ ] Command parsing logic implemented

- [ ]  Secure handling of the password

- [ ]   Linked to Issue #37

feature is not intended to broadcast the "this channel has password"  to everyone . Open to feedback on business rules here.

### DIFF
--- a/app/src/main/java/com/bitchat/android/ui/ChannelManager.kt
+++ b/app/src/main/java/com/bitchat/android/ui/ChannelManager.kt
@@ -284,9 +284,21 @@ class ChannelManager(
         state.setShowPasswordPrompt(false)
         state.setPasswordPromptChannel(null)
     }
-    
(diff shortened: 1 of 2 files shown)


callebtc:
thank you!

Issue #7, opened 2025-07-09
Translate project


hosseinabaspanah:
Hello, thank you for implementing this idea and for your concern for user privacy.
I am excited to announce my readiness to translate this project to Luri Bakhtiari (bqi) and I am happy to be among the first to do the translation work for this project.

Where should I start?

Bobcat5360:
Translate to what?

hosseinabaspanah:
> Translate to what?

Luri Bakhtiari

hosseinabaspanah:
@Bobcat5360 Please check this link:
https://en.m.wikipedia.org/wiki/Bakhtiari_dialect

SongOnWater:
https://developer.android.com/guide/topics/resources/localization

callebtc:
would love translations built in!

Bobcat5360:
#40

Bobcat5360:
English to French

Bobcat5360:
#98

Bobcat5360:
close?

Bobcat5360:
#66 #25

hosseinabaspanah:
I have now prepared the translation file for Luri Bakhtiari.
Where should I send it?

Bobcat5360:
@callebtc

PR #40, opened 2025-07-10
provide translatable strings for the permission explanation screen

vashisthg:
# Description
Puts the strings in the translatable strings.xml 

A part of #7 

## Checklist


app/src/main/java/com/bitchat/android/onboarding/PermissionExplanationScreen.kt
```diff
@@ -94,10 +96,10 @@ fun PermissionExplanationScreen(
                 }
                 
                 Text(
-                    text = "• bitchat doesn't track you or collect personal data\n" +
```
vladwing: This is the wrong way to do translations because those splits might not make sense in other languages.

app/src/main/java/com/bitchat/android/onboarding/PermissionExplanationScreen.kt
```diff
@@ -237,20 +239,22 @@ private fun PermissionCategoryCard(
     }
 }
 
+@Composable
 private fun getPermissionEmoji(categoryName: String): String {
     return when (categoryName) {
-        "Nearby Devices" -> "📱"
```
vladwing: This fails on right to left languages for example.

app/src/main/java/com/bitchat/android/onboarding/PermissionExplanationScreen.kt
vashisthg: You are right

app/src/main/java/com/bitchat/android/onboarding/PermissionExplanationScreen.kt
vashisthg: Can you explain how. This is just a string comparison.

vashisthg:
I see #78 has some similar changes too regarding adding an enum

app/src/main/java/com/bitchat/android/onboarding/PermissionExplanationScreen.kt
vashisthg: This line itself has been removed now

Chessing234 (commented):
several of the new strings bake a leading quote into the value (box2_title, the grant/exit buttons, the permission titles). that will render as "Grant Permissions". also box1_text uses literal \n — use a real newline or a separate paragraph, not escaped n.

Issue #52, opened 2025-07-11
APK version does not match release version


rollingmoai:
### Checklist

- [x] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/callebtc/bitchat-android/releases/latest).
- [x] I made sure that there are *no existing issues* - [open](https://github.com/callebtc/bitchat-android/issues) or [closed](https://github.com/callebtc/bitchat-android/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] This issue contains only one bug.
- [x] I have read and understood the [contribution guidelines](https://github.com/callebtc/bitchat-android/blob/main/README.md#contributing).

### Affected version

0.5.1

### Steps to reproduce the bug

1. Install BitChat
2. Go to app info to check app version

### Expected behavior

App version matches release version (`0.5.1`)

### Actual behavior

App version does not match release version

### Screenshots/Screen recordings

![Image](https://github.com/user-attachments/assets/9b7e477d-2576-4283-b9e7-fc24c7d2ff8a)

rollingmoai:
This seems to be fixed now as of https://github.com/permissionlesstech/bitchat-android/releases/tag/0.6

<img width="1080" height="574" alt="Image" src="https://github.com/user-attachments/assets/c0b4b6e0-6713-4d13-b7d9-7a40c40e4259" />

PR #54, opened 2025-07-11
Automatic versionName and versionCode generation from git tags

mustfaunlu:
# Description
Automatic versionName and versionCode assignment has been implemented using the git tag and a versionCode derived from versionName. Now, when a new release tag is created and pushed, versionName and versionCode will be set automatically.
This eliminates the need for manual version updates and ensures consistent version management with CI/CD.
While this approach is a practical and effective solution for now, it can be replaced with a more robust or advanced versioning strategy in the future if project requirements evolve.       
Closes: https://github.com/permissionlesstech/bitchat-android/issues/52                       
## Checklist

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,3 +1,5 @@
+import java.io.ByteArrayOutputStream
+
 plugins {
     alias(libs.plugins.android.application)
     alias(libs.plugins.kotlin.android)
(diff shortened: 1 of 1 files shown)


vashisthg:
Won't this slow down the debug builds too?

mustfaunlu:
Yes, running a git command during every build can cause a slight slowdown in debug builds. However, since these commands are usually very fast, the slowdown is generally not noticeable in small to medium-sized projects. In very large projects or when build speed is critical, this can be optimized in the future to run only for release builds or by using a different method.

vladwing:
There are a few side effects of using tags for versions, which are not obvious, but they caused issues for me in the past:
1. `git describe` doesn't look only on the current commit, if it doesn't have a tag, but looks back in history for the most recent tag with a certain format. This can prove problematic in CI/CD because they only pull very few commits to optimize the checkout. If the tag is too far back in history (not such an outlandish possibility if the checkout is done with 'git clone --shallow=1) `git describe` will not return the correct version.
2. tags can be deleted, moved, etc. That means a build  might get different results (because the version is different) when executed on the same hash. It breaks a lot of sanity assumptions.
3. Tags are many times used to represent a certain quality stamp on a build (eh. passed certain checks), hence they should be done after the build not before.

My suggestion is to check-in the version in the repo and the maintainer can update the version when they see fit.

mustfaunlu:
issue completed

PR #57, opened 2025-07-11
Cheap routing optimizations

GUVWAF:
Sending a packet is currently implemented by sequentially sending to each connected peripheral or central. We can reduce airtime and battery usage without requiring any control packets to be sent. Since Bitchat relies on the connection-oriented mode of BLE, we already have knowledge about our 0-hop neighborhood. After a key exchange we can map a Bluetooth address to a Bitchat `peerID` for a bit more intelligent routing.

This PR makes sure to:

1. Not relay a packet to the peripheral/central matching the `senderID` or the relayer of this packet, as they already generated or received the packet. While these will be ignored by the receiver, let’s not waste airtime on this in the first place.
2. Send a packet with specific destination only to the peripheral/central matching the `recipientID`, if we’re directly connected. This potentially avoids a lot of unnecessary rebroadcasting.
3. Not relay an ACK or read receipt coming from ourselves (same was already done for messages).

Most important changes are in `broadcastPacket()`. I introduced a new `addressPeerMap` to map Bluetooth addresses to `peerID`s, and a `RoutedPacket` data structure containing a `BitchatPacket` and routing metadata.

Tested with Android devices. It’s backwards compatible with both Android and iOS since no new packets or header entries are needed, it only mitigates unnecessary transmissions. Since this is a protocol-level optimization, I’ll open an issue requesting these optimizations for iOS also.


callebtc:
Awesome, thank you! Can you expand on point 2? Shouldn't we relay private messages even if the sender and recipient don't have a direct connection? Maybe I'm misunderstanding

GUVWAF:
> Shouldn't we relay private messages even if the sender and recipient don't have a direct connection? 

Yes, indeed. So in `broadcastPacket()` it now first checks *if* we have a direct connection. If so, we send directly and do an early return. If not, we'll continue with the normal broadcasting to all peripherals/centrals.

GUVWAF:
That reminded me that the direct send may fail (e.g. connection was lost in the meantime), so we should also fall back to broadcast in that case. Latest commit ensures that.

callebtc:
Best PR so far. Thank you.

PR #59, opened 2025-07-11
release mode to be minified and shrinked

prudhvir3ddy:
# Description

Why are we doing this? 

https://developer.android.com/topic/performance/app-optimization/enable-app-optimization

>To enable app optimization, set isMinifyEnabled = true (for code optimization) and isShrinkResources = true (for resource optimization) in your [release build's](https://developer.android.com/studio/publish/preparing#turn-off-debugging) app-level build script as shown in the following code. We recommend that you always enable both settings. We also recommend enabling app optimization only in the final version of your app that you test before publishing—usually your release build—because the optimizations increase the build time of your project and can make debugging harder due to the way it modifies code.

Here's the working. 

https://github.com/user-attachments/assets/707f5a69-82ad-4193-a4bb-3c431f67bf7a

## Checklist

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -23,7 +23,8 @@ android {
 
     buildTypes {
         release {
-            isMinifyEnabled = false
+            isMinifyEnabled = true
+            isShrinkResources = true
             proguardFiles(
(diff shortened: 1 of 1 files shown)


callebtc:
you're the best

PR #64, opened 2025-07-11
Optimize PNG images losslessly: 12.4 → 10.2 KiB (−18%)

maximal:


callebtc:
dayum we don't use these logos, they are a leftover from my ios port :/

callebtc:
how about we go to 0 KiB? :)

prudhvir3ddy:
@callebtc We are using these logos. 

That's how the app icon shows up for different android devices.

https://github.com/permissionlesstech/bitchat-android/blob/main/app/src/main/AndroidManifest.xml#L31-L32

callebtc:
Ok thanks, the let's use these and get rid of the SVG logo! Would you adjust your PR so we use the PNG icon exclusively? It would be great if it also works well with circular and adaptive color icons for Android. I'd truly appreciate your help with this.

callebtc:
Ok thanks, then let's use these and get rid of the SVG logo! 

Would you adjust your PR so we use the PNG icon exclusively? It would be great if it also works well with circular and adaptive color icons for Android. I'd truly appreciate your help with this.

PR #86, opened 2025-07-12
Add Robolectric for testing Android specific code

shubertm:
# Description
Adds Robolectric 4.15.1 and unit tests for some parts of EncryptionService

## Checklist


app/src/androidTest/kotlin/com/bitchat/EncryptionServiceTest.kt
```diff
@@ -0,0 +1,30 @@
...
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import com.bitchat.android.crypto.EncryptionService
+import com.bitchat.android.crypto.EncryptionService.Companion.PUBLIC_KEY_DATA_SIZE
+import junit.framework.TestCase.assertEquals
+import junit.framework.TestCase.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
```
prudhvir3ddy: I'm not sure why we need an instrumented test for this case. 

You can just mock or fake the context and write a junit tests 

> [!NOTE]
>Junit tests are much faster than automated UI tests. 
Write automated tests for verifying the behaviour. For logics and all  junit tests that doesn't involve a device is great.

prudhvir3ddy (changes requested):
For instrumented tests, ensure a UI check or system-level APIs check is happening

app/src/androidTest/kotlin/com/bitchat/EncryptionServiceTest.kt
shubertm: I was looking for the context, I think mocking is a good idea. I will refactor it to unit tests

callebtc:
I'm not sure if it's worth adding a test suite dependency with only a simple test file. what's the benefit of this PR?

shubertm:
> I'm not sure if it's worth adding a test suite dependency with only a simple test file. what's the benefit of this PR?

I want to write tests for the encryption service, found out it requires context, so the first way was to use instrumented tests but they are too slow. @prudhvir3ddy suggests unit tests with mocked context. I am not sure if we can mock without any library like robolectric or mockito.

prudhvir3ddy:
@shubertm 

Reason this class requires a context is because of shared prefs 

you can pass shared prefs as a dependency to this class and pass fake prefs from the tests.

shubertm:
> @shubertm
> 
> Reason this class requires a context is because of shared prefs
> 
> you can pass shared prefs as a dependency to this class and pass fake prefs from the tests.

@prudhvir3ddy How can you add the fake prefs without library deps? I been looking it up, found nothing

PR #102, opened 2025-07-13
Revert "known issues added to 0.6"

callebtc:
Reverts permissionlesstech/bitchat-android#93


Comodore125:
What was the issue with knows issues addition?

callebtc:
The issue is not tied to any specific version. I need to investigate if it's fixed now or still persist.

Comodore125:
thanks for clarification. it helps tremendously.

PR #103, opened 2025-07-13
adjust scan intervals and reduce logging

callebtc:
# Description

## Checklist

### DIFF
--- a/app/src/main/java/com/bitchat/android/MainActivity.kt
+++ b/app/src/main/java/com/bitchat/android/MainActivity.kt
@@ -33,6 +33,7 @@ class MainActivity : ComponentActivity() {
     private lateinit var permissionManager: PermissionManager
     private lateinit var onboardingCoordinator: OnboardingCoordinator
     private lateinit var bluetoothStatusManager: BluetoothStatusManager
+    private lateinit var locationStatusManager: LocationStatusManager
     
     // Core mesh service - managed at app level
     private lateinit var meshService: BluetoothMeshService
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/bitchat/android/MainActivity.kt
```diff
@@ -189,7 +189,7 @@ class MainActivity : ComponentActivity() {
      * Check Bluetooth status and proceed with onboarding flow
      */
     private fun checkBluetoothAndProceed() {
-        android.util.Log.d("MainActivity", "Checking Bluetooth status")
+        // android.util.Log.d("MainActivity", "Checking Bluetooth status")
```
prudhvir3ddy, line 216: Can delete it instead of comment.

app/src/main/java/com/bitchat/android/mesh/BluetoothConnectionManager.kt
```diff
@@ -148,7 +148,7 @@ class BluetoothConnectionManager(
                 
                 startPeriodicCleanup()
                 
-                Log.i(TAG, "Power-optimized Bluetooth services started successfully")
+                Log.i(TAG, "Bluetooth services started successfully")
```
prudhvir3ddy, line 154: > [!TIP]
> For logging, you can use a library like [Timber](https://github.com/JakeWharton/timber), which helps you with 
> 1. Not required to add a Tag. It automatically picks the class name as a tag in which it ran 
> 2. Not calling the logging code in production

PR #139, opened 2025-07-16
refactor: remove context attribute at ChatViewModel.kt

alpermelkeli:
# Description

## Checklist

- [+] I have read the contribution guidelines: <https://github.com/permissionlesstech/bitchat-android?tab=readme-ov-file#contributing>
- [+] I have performed a self-review of my code

- [+] I have mentioned the corresponding issue and the relevant keyword (e.g., "Closes: #xy") in the description (see <https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue>)
- [+] If it is a core feature, I have added automated tests

Store context as a static field inside of classes can cause memory leaks. 
`
  private val context: Context = application.applicationContext
`

### DIFF
--- a/app/src/main/java/com/bitchat/android/ui/ChatViewModel.kt
+++ b/app/src/main/java/com/bitchat/android/ui/ChatViewModel.kt
@@ -24,14 +24,12 @@ class ChatViewModel(
     application: Application,
     val meshService: BluetoothMeshService
 ) : AndroidViewModel(application), BluetoothMeshDelegate {
-    
-    private val context: Context = application.applicationContext
-    
+
     // State management
     private val state = ChatState()
     
     // Specialized managers
-    private val dataManager = DataManager(context)
+    private val dataManager = DataManager(application.applicationContext)
     private val messageManager = MessageManager(state)
     private val channelManager = ChannelManager(state, messageManager, dataManager, viewModelScope)
     val privateChatManager = PrivateChatManager(state, messageManager, dataManager)
@@ -46,7 +44,7 @@ class ChatViewModel(
         privateChatManager = privateChatManager,
         notificationManager = notificationManager,
         coroutineScope = viewModelScope,
-        onHapticFeedback = { ChatViewModelUtils.triggerHapticFeedback(context) },
+        onHapticFeedback = { ChatViewModelUtils.triggerHapticFeedback(application.applicationContext) },
         getMyPeerID = { meshService.myPeerID }
     )
     


moehamade:
@callebtc No memory leak will ever happen here since we're using application context not activity or ui related context, so this application context will stay alive as long as the app is running outliving every ViewModel in the app, however Android Studio flags this warning as potential memory leak, we could either:
1- Suppress the warning
2- Accept the PR since it might look cleaner to not hold application context in a context variable

But I just wanted to highlight that no memory leak will ever happen here

Side note: Managers could benefits from some dependency injection either manual or using a library instead of having to create them each time and pass in their related dependencies.

alpermelkeli:
> @callebtc No memory leak will ever happen here since we're using application context not activity or ui related context, so this application context will stay alive as long as the app is running outliving every ViewModel in the app, however Android Studio flags this warning as potential memory leak, we could either: 1- Suppress the warning 2- Accept the PR since it might look cleaner to not hold application context in a context variable
> 
> But I just wanted to highlight that no memory leak will ever happen here
> 
> Side note: Managers could benefits from some dependency injection either manual or using a library instead of having to create them each time and pass in their related dependencies.

We can use ktor I think.

moehamade:
> > @callebtc No memory leak will ever happen here since we're using application context not activity or ui related context, so this application context will stay alive as long as the app is running outliving every ViewModel in the app, however Android Studio flags this warning as potential memory leak, we could either: 1- Suppress the warning 2- Accept the PR since it might look cleaner to not hold application context in a context variable
> > But I just wanted to highlight that no memory leak will ever happen here
> > Side note: Managers could benefits from some dependency injection either manual or using a library instead of having to create them each time and pass in their related dependencies.
> 
> We can use ktor I think.

ktor is for network requests, I think you're referring to koin [https://insert-koin.io/docs/quickstart/android/](url)

alpermelkeli:
> > > @callebtc No memory leak will ever happen here since we're using application context not activity or ui related context, so this application context will stay alive as long as the app is running outliving every ViewModel in the app, however Android Studio flags this warning as potential memory leak, we could either: 1- Suppress the warning 2- Accept the PR since it might look cleaner to not hold application context in a context variable
> > > But I just wanted to highlight that no memory leak will ever happen here
> > > Side note: Managers could benefits from some dependency injection either manual or using a library instead of having to create them each time and pass in their related dependencies.
> > 
> > 
> > We can use ktor I think.
> 
> ktor is for network requests, I think you're referring to koin [https://insert-koin.io/docs/quickstart/android/](url)

yeah sorry I mean koin.

callebtc:
Thank you for your PR! appreciated

Issue #141, opened 2025-07-16
Toolbar height looks off; center content vertically if intended


hector6872:
### Checklist

- [x] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/permissionlesstech/bitchat-android/releases/latest).
- [x] I made sure that there are *no existing issues* - [open](https://github.com/permissionlesstech/bitchat-android/issues) or [closed](https://github.com/permissionlesstech/bitchat-android/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] This issue contains only one bug.
- [x] I have read and understood the [contribution guidelines](https://github.com/permissionlesstech/bitchat-android/blob/main/README.md#contributing).

### Affected version

0.7

### Steps to reproduce the bug

I'm not sure whether this is by design, but the toolbar height looks off. 
If this is the intended size, we should at least center the content vertically.

### Screenshots/Screen recordings

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/d2b3c11b-8750-4bbb-9967-b1f33b0cbc27" />

### Affected Android/Custom ROM version

Android 12/16

PR #142, opened 2025-07-16
Fix content padding in messages list

hector6872:
# Description

| Before | After |
|--------|--------|
| before | after | 

## Checklist


moehamade:
I think current approach was better (especially that the header text will be centered after #141 is merged, so we will have 2 texts close to each other).
The better solution would be to make the padding container a gradient with low opacity so it feels an invisible white layer in light mode that will appear slightly faded when an item from the list is behind it.

hector6872:
> I think current approach was better

The current method chops the text and it doesn't look good IMO

> The better solution would be to make the padding container a gradient with low opacity

Easy, but very ListView-era. Modern apps don't do this.

callebtc:
Thank you, great fix!

PR #120, opened 2025-07-14
Fix: wrong cursor location, added battery optimization disabling screen and styling to the slash command in input field

anonfaded:
# Description
- Now the cursor will auto move ahead of the slash command when we select a slash command from the list. 
- Added some styling to the command so it looks better and doesn't mix with other text in input field.
- Added battery optimization permission and management system to ensure app runs reliably in background.

## Screenshots

image

image

image

Closes: #119 

## Checklist

- [x] I have read the contribution guidelines: <https://github.com/permissionlesstech/bitchat-android?tab=readme-ov-file#contributing>
- [x] I have performed a self-review of my code

- [x] I have mentioned the corresponding issue and the relevant keyword (e.g., "Closes: #xy") in the description (see <https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue>)

PR #144, opened 2025-07-17
Move cursor to end on Command selection

### LINKED ISSUE #119: [BUG] Cursor is not placed ahead of slash command
### Checklist

### Affected version

0.7

Type `/` and select any command from the list and the cursor will still be placed at the place of slash (`/`) instead of appearing ahead of it.

### Additional information

I already fixed this so please check the linked PR. Also added some styling to the slash command to make it identical in input field

hector6872:
# Description

Closes #119

| Before | After |
|--------|--------|
| Screenshot 2025-07-17 at 10 56 10 | Screenshot 2025-07-17 at 10 55 10 | 

## Checklist

### DIFF
--- a/app/src/main/java/com/bitchat/android/ui/ChatScreen.kt
+++ b/app/src/main/java/com/bitchat/android/ui/ChatScreen.kt
@@ -1,6 +1,5 @@
 package com.bitchat.android.ui
 
-import android.annotation.SuppressLint
 import androidx.compose.animation.*
 import androidx.compose.animation.core.*
 import androidx.compose.foundation.*
@@ -26,7 +25,6 @@ import androidx.compose.ui.zIndex
  * - DialogComponents: Password prompts and modals
  * - ChatUIUtils: Utility functions for formatting and colors
  */
-@SuppressLint("UnusedBoxWithConstraintsScope")
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun ChatScreen(viewModel: ChatViewModel) {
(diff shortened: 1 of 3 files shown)


moehamade:
@callebtc Slash command is also fixed in issue #120 but that issue is too big and introduces multiple stuff all in 1 PR, 
so maybe merging this would be better.
@hector6872 but just 1 thing to note, I think isEmpty() doesn't account for whitespaces, so try checking if you should use isBlank() instead of isEmpty()

hector6872:
@moehamade I used the Kotlin extension since it basically does the same thing in that case :) 

```
/**
 * Returns `true` if this char sequence is empty (contains no characters).
 *
 * @sample samples.text.Strings.stringIsEmpty
 */
@kotlin.internal.InlineOnly
public inline fun CharSequence.isEmpty(): Boolean = length == 0

```

Checking if `length == 0` is the same as checking if it's `length < 1`.

moehamade:
@hector6872 hey hector, I just tested it if we removed this extra condition completely it has no effect, what we're basically saying if input doesn't start with '/' dismiss the commands suggestions, so this applies if it's whitespace (white space is not '/' , empty (empty char is not '/') so  no need for that extra condition. Otherwise great work!

hector6872:
@moehamade given limited test coverage and uncertain impact, I restricted the change to minor Kotlin-idiomatic cleanup but if you tested it, I'm fine changing it :)

callebtc:
Appreciate the PR! however I think this is a dupe of #120

callebtc:
sorry that this one was duped, I hope we'll see more PRs from you!

PR #146, opened 2025-07-17
Add CommandSuggestionsBox quick access

hector6872:
# Description

Add quick access to show or hide the `CommandSuggestionBox`. We aim to make the UI more user-friendly to attract more users.

Screen recording attached.

| Before | After |
|--------|--------|
| Screenshot_20250717_184704 | Screen_recording_20250717_184510 | 

## Checklist


callebtc:
I like this! what do you think of making it a button with a light fill so it looks a bit less like a letter in the input field?

hector6872:
Sure! I'll update the PR ASAP

hector6872:
@callebtc Material3 provides a FilledTonalIconButton, which seems like the right fit here :)

> A filled tonal icon button is a medium-emphasis icon button that is an alternative middle ground between the default FilledIconButton and OutlinedIconButton. They can be used in contexts where the lower-priority icon button requires slightly more emphasis than an outline would give.

image
Screenshot 2025-07-21 at 00 11 01

callebtc:
@hector6872 thank you for this and sorry for the delay!

callebtc:
I've replaced the send button with the quick command button if the text field is empty. I think we can use that button in the future to access more features.

callebtc:
@hector6872 please keep it coming, I could use some help! :)

PR #155, opened 2025-07-19
Add Italian localization

riccio8:
Add Italian localization (values-it/strings.xml)

# Description

## Checklist


Chessing234 (commented):
this looks superseded — `app/src/main/res/values-it/strings.xml` is on main with 359 translated keys.

this branch carries 34 keys, and all 34 are already translated on main-it. so the diff is a no-op against current main.

the italian catalog isn't finished though: 204 keys in `values/strings.xml` still have no it translation. if you're still up for it, a fresh branch against main filling those in would be genuinely useful — that's where the gap is now.

PR #159, opened 2025-07-20
Enhancement: Added Support for multi-line text in Chat Input

### LINKED ISSUE #157: [Request] Enter key sends option.
### Checklist

### App version

0.7.1

Pretty simple feature, just want to be able to toggle on the ability to send messages with the enter key. Additionally, this toggle could also set shift+enter, to add a new line, as the enter key currently does.

### Why should this be added?

I use a phone with a hardware keyboard, and it's just inconvenient to have to send messages with the on-screen button. Lots of other messaging apps have this as either a toggleable option, or the default behaviour of the enter key. Both Discord and Signal come to mind as examples of applications that have this as their default behaviour. I don't think its a huge ask as it shouldn't be too difficult to implement, and it would make the app significantly more comfortable to use for people with hardware keyboards.

_No response_

moehamade:
Closes #157 
Added Support for multi line text
The IME action for the text input field has been changed from `ImeAction.Send` to `ImeAction.Default`. This ensures the keyboard displays the appropriate action based on the context, rather than always showing a "Send" button. -In this context, default will be to go to a new line instead of send, (there is no other way to go to a new line)

# Description

- Added line break utility function for TextFieldValue
- Added support for multi line text
- Added special case support for external keyboard (Enter: Send message / Shift + Enter: Go to new line)

## Checklist

### DIFF
--- a/app/src/main/java/com/bitchat/android/core/ui/utils/TextFieldValueExt.kt
+++ b/app/src/main/java/com/bitchat/android/core/ui/utils/TextFieldValueExt.kt
@@ -20,7 +20,6 @@ import androidx.compose.ui.text.input.TextFieldValue
 fun TextFieldValue.lineBreak(
     onValueChange: (TextFieldValue) -> Unit
 ) {
-    val selection = this.selection
     val newTextValue = TextFieldValue(
         text = this.text.substring(
             0,
--- a/app/src/main/java/com/bitchat/android/core/ui/utils/TextFieldValueExt.kt
+++ b/app/src/main/java/com/bitchat/android/core/ui/utils/TextFieldValueExt.kt
@@ -1,0 +1,32 @@
+package com.bitchat.android.core.ui.utils
+
+import androidx.compose.ui.text.TextRange
+import androidx.compose.ui.text.input.TextFieldValue
+
+/**
+ * Inserts a line break (`\n`) into the [TextFieldValue] at the current cursor position or replaces the selected text with a line break.
+ *
+ * This function modifies the text by:
+ * 1. Taking the substring before the current selection start.
+ * 2. Appending a newline character (`\n`).
+ * 3. Appending the substring after the current selection end.
+ *
+ * The cursor is then moved to the position immediately after the inserted newline character.
+ *
+ * @param onValueChange A lambda function that will be invoked with the new [TextFieldValue]
+ *                      after the line break has been inserted. This is typically used to update
+ *                      the state of a `TextField`.
+ */
+fun TextFieldValue.lineBreak(
+    onValueChange: (TextFieldValue) -> Unit
+) {
+    val selection = this.selection
+    val newTextValue = TextFieldValue(
+        text = this.text.substring(
+            0,
+            selection.start
+        ) + '\n' + this.text.substring(selection.end),
+        selection = TextRange(selection.start + 1)
+    )
+    onValueChange(newTextValue)
+}
--- a/app/src/main/java/com/bitchat/android/ui/InputComponents.kt
+++ b/app/src/main/java/com/bitchat/android/ui/InputComponents.kt
@@ -1,6 +1,5 @@
 package com.bitchat.android.ui
 
-import android.util.Log
 import androidx.compose.foundation.background
 import androidx.compose.foundation.border
 import androidx.compose.foundation.clickable
@@ -39,7 +38,6 @@ import androidx.compose.ui.input.key.onPreInterceptKeyBeforeSoftKeyboard
 import androidx.compose.ui.input.key.type
 import androidx.compose.ui.text.AnnotatedString
 import androidx.compose.ui.text.SpanStyle
-import androidx.compose.ui.text.TextRange
 import androidx.compose.ui.text.buildAnnotatedString
 import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.font.FontWeight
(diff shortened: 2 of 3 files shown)


moehamade:
@callebtc Hey, I made default keyboard done button to be go to new line instead of send message since there was no way to go to a new line without an external keyboard, however note that iOS also has no way to go to a new line as of now.

app/src/main/java/com/bitchat/android/core/ui/utils/TextFieldValueExt.kt
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @param onValueChange A lambda function that will be invoked with the new [TextFieldValue]
+ *                      after the line break has been inserted. This is typically used to update
+ *                      the state of a `TextField`.
+ */
+fun TextFieldValue.lineBreak(
+    onValueChange: (TextFieldValue) -> Unit
+) {
+    val newTextValue = TextFieldValue(
+        text = this.text.substring(
```
hector6872, line 24: We can use this instead to avoid clutter with multiple substring and concat calls:

`StringBuilder(this.text).insert(selection.start, "\n").toString()`

app/src/main/java/com/bitchat/android/ui/InputComponents.kt
```diff
@@ -125,13 +146,25 @@ fun MessageInput(
...
             cursorBrush = SolidColor(colorScheme.primary),
-            keyboardOptions = KeyboardOptions(imeAction = ImeAction.Send),
+            keyboardOptions = KeyboardOptions(imeAction = ImeAction.Default),
             keyboardActions = KeyboardActions(onSend = { onSend() }),
             visualTransformation = SlashCommandVisualTransformation(),
             modifier = Modifier
                 .weight(1f)
-                .onFocusChanged { focusState ->
-                    isFocused.value = focusState.isFocused
+                .onFocusChanged { focusState -> isFocused.value = focusState.isFocused }
```
hector6872, line 154: What’s this for? Could we remove it?

app/src/main/java/com/bitchat/android/ui/InputComponents.kt
```diff
@@ -125,13 +146,25 @@ fun MessageInput(
...
+                .onPreInterceptKeyBeforeSoftKeyboard { keyEvent ->
+                    if (keyEvent.isShiftPressed) {
```
hector6872, line 156: This if statement could be simplified:
```
if (keyEvent.key == Key.Enter || keyEvent.key == Key.NumPadEnter) {
    when {
        keyEvent.isShiftPressed && keyEvent.type == KeyEventType.KeyDown -> value.lineBreak(onValueChange)
        else -> onSend()
    }
    return@onPreInterceptKeyBeforeSoftKeyboard true
}
false
```

What's the reason for evaluating `keyEvent.type == KeyEventType.KeyDown`?

moehamade: Hello Hector!, thanks for the review, I actually took some time to figure out this issue, in Android there are 4 ways to intercept key events,  [https://developer.android.com/develop/ui/compose/touch-input/keyboard-input/commands](url).

1. onKeyEvent()
2. onPreviewKeyEvent()
3. onInterceptKeyEvent()
4. onPreInterceptKeyEvent()

and in each case we should handle 2 types of key events, the pressing (KeyUp) and release (KeyDown),
it was causing me to trigger event twice because 1 is executed on KeyUp and 1 for keyDown which will trigger it twice. So yea I am only interested in KeyDown in this case. Hope this answers ur question!

app/src/main/java/com/bitchat/android/ui/InputComponents.kt
moehamade: tbh I don't think it's bad to have this on 1 line since it's small and onKeyEvent that I added is way too big, I preferred to have one liners before the onKeyEvent so it's more readable, I think we should have clear maintaining rules and conventions so we don't end up with such stuff, but for now I think it's fine

app/src/main/java/com/bitchat/android/core/ui/utils/TextFieldValueExt.kt
moehamade: Thanks for the suggestion! Each one has its pros and cons, but in this case I think using stringbuilder is overhead for messages, the only minor consideration is that if you were doing multiple string manipulations, StringBuilder might be worth it, but for a single line break insertion, the direct string concatenation approach might work better.

On a side note, with ur approach if we select the message in text box, and then decide to shift + enter to go to a new line, it won't clear the string and will place the cursor at start, in my approach it will clear the selection (desired approach) and will place the cursor in the correct position. Let me know if that answers ur concerns!

app/src/main/java/com/bitchat/android/ui/InputComponents.kt
hector6872: Thanks for taking the time to clarify this! 🤝

app/src/main/java/com/bitchat/android/ui/InputComponents.kt
hector6872: I'm fine with the one-liner (it looks better IMO). But we're not using the `isFocused` var anywhere.

moehamade: Oh I get what u mean now, usually I don’t like changing logic not related to PR, maybe we can do this in a separate one, or keep it for now in case they initially intended to do something else, up to @callebtc

Chessing234 (changes requested):
lineBreak indexes selection.start/end with no ordering check — a reversed selection throws. also ImeAction.Default means the leftover onSend keyboardaction never fires, so send is button-only now; please say if that's intended.

PR #166, opened 2025-07-21
Auto-generate versionCode from versionName

hector6872:
# Description

Improves app version management by using a single .toml source for `versionName` and auto-calculating `versionCode` from it.

| versionName  | versionCode  |
|---|---|
| 0.7.2 | 7002  |
| 0.7.3 | 7003  |
| 1.0.80 | 1000080 |
| 1.8.2 | 1008002 |
| 1.12.0 | 1012000 |

## Checklist

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -55,18 +55,17 @@ android {
 }
 
 
-private fun getVersionName(): String = app.versions.major.get() +
-    ".${app.versions.minor.get()}" +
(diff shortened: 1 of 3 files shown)


moehamade (commented):
I also think that these should be automated, but I'm not sure if we should just add 0s or do it in a better way, but I think for now, since we don't have any release on Playstore, (whether internal testing or release), it's too early. Optimally those should be set during continuous integration I think

PR #176, opened 2025-07-23
Update README.md - Mention Bitchat-tui

Comodore125:
Mention of Bitchat-tui in readme as it is another implementation of bitchat while being terminal based tool built in Rust for Windows/Linux/MacOS environements

# Description
I believe this enhances readme.md but lets see what maintaner thinks. For me, this bitchat-tui was a great find as I prefer to do stuff on computer instead of phone. 
## Checklist


Comodore125:
Maybe, this PR could wait, Bitchat tui development is not at the level of those two main bitchats.

Comodore125:
Currently there is no development ongoing on bitchat-tui. will update in couple of weeks and close if it dies.

callebtc:
We can do this if the project looks like it keeps moving.

PR #198, opened 2025-07-30
improve identity announcements

callebtc:
# Description

## Checklist


GmrsKissimmee:
Global Bluetooth Mesh

Is it possible to make it use all nearby BLE devices as support and thus create a super mesh with everything around it?

Node, mesh, relay (smart locks, cameras, Amazon access control in each neighborhood, Bluetooth repeaters in warehouses, cars, cell phones, networks, Meshtastic nodes, etc.)

Just use the BLE point for the network, not interfere with each of them, as a detection-only protocol and information hopping.

Issue #200, opened 2025-07-30
When you change your nickname, the sender color of your past messages also changes.


whflima:
### Checklist

- [x] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/permissionlesstech/bitchat-android/releases/latest).
- [x] I made sure that there are *no existing issues* - [open](https://github.com/permissionlesstech/bitchat-android/issues) or [closed](https://github.com/permissionlesstech/bitchat-android/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] This issue contains only one bug.
- [x] I have read and understood the [contribution guidelines](https://github.com/permissionlesstech/bitchat-android/blob/main/README.md#contributing).

### Affected version

0.8

### Steps to reproduce the bug

1. Open the app.
2. Send a few messages on the main screen.
3. Change your nickname.
4. Observe that the sender color of your previous messages has also changed.
5. On another device, set a user’s nickname to match yours exactly.
6. Notice that both users now share the same sender color. 

### Expected behavior

The sender color of previous messages should remain the same, even after changing your nickname.

or

When the nickname is changed, previous messages should display the new nickname but retain their original sender color. (More complex)

### Screenshots/Screen recordings

1. Single user: changing nickname updates sender color in past messages.

https://github.com/user-attachments/assets/e5394255-da32-469a-beb2-e93ce429a873

2. When two or more users use the same nickname.

https://github.com/user-attachments/assets/9fce2ba5-a1d6-4d34-9b8e-da24d6b691d9

### Affected Android/Custom ROM version

Android 16.0

PR #201, opened 2025-07-30
fix(ui): prevent sender color of past messages

whflima:
Closes https://github.com/permissionlesstech/bitchat-android/issues/200

# Description

Previously, the sender was identified by comparing `message.sender` (a nickname string) to `currentUserNickname`. This caused issues when the nickname changed, since past messages would then match or not match incorrectly, leading to inconsistent color assignments.

The updated logic compares `message.senderPeerID` to `meshService.myPeerID`, which is a unique and stable identifier for the user. This ensures that the correct color (`colorScheme.primary`) is applied consistently to the user's own messages, regardless of any nickname changes.

# Why this matters

- Fixes a bug where changing your nickname would retroactively change the color of your past messages.

- Prevents conflicts where multiple users with the same nickname might get the same sender color.

- Makes message ownership checks more reliable by using a unique ID instead of a mutable nickname.

# Screenshots/Screen recordings

https://github.com/user-attachments/assets/c53e2dbd-d652-4c3a-b9a3-19a27364dde1

## Checklist

### DIFF
--- a/app/src/main/java/com/bitchat/android/ui/ChatUIUtils.kt
+++ b/app/src/main/java/com/bitchat/android/ui/ChatUIUtils.kt
@@ -54,7 +54,7 @@ fun formatMessageAsAnnotatedString(
     if (message.sender != "system") {
         // Sender
         val senderColor = when {
-            message.sender == currentUserNickname -> colorScheme.primary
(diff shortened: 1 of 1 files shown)


callebtc:
nice catch, thank you

Issue #131, opened 2025-07-15
We can send message without username


amitthecoder:
### Checklist

- [ ] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/permissionlesstech/bitchat-android/releases/latest).
- [x] I made sure that there are *no existing issues* - [open](https://github.com/permissionlesstech/bitchat-android/issues) or [closed](https://github.com/permissionlesstech/bitchat-android/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] This issue contains only one bug.
- [x] I have read and understood the [contribution guidelines](https://github.com/permissionlesstech/bitchat-android/blob/main/README.md#contributing).

### Affected version

We see in latest version

### Steps to reproduce the bug

We send message to another user without username that hard to know source of message

moehamade:
@callebtc this is iOS behavior as well, not sure if we should change it on Android

twlgsc:
The same problem occurs on Android devices, at least the Samsung Galaxy S24 Ultra.

BEIMYT:
> The same problem occurs on Android devices, at least the Samsung Galaxy S24 Ultra.

Also happens in my Samsung Galaxy A13 4G

Tsurgcom:
Same here on:
* Samsung S25
* Samsung S21 5G
* Samsung A14

PR #207, opened 2025-07-31
fix: generate new nickname if NicknameEditor loses focus with blank value

whflima:
Closes https://github.com/permissionlesstech/bitchat-android/issues/131

# Description

This PR updates the `NicknameEditor` behavior to automatically generate a new nickname when the input `BasicTextField` loses focus and the current value is blank. This ensures that users are always assigned a nickname if they leave the field black.

# Why this matters

- Prevents users from leaving the nickname field empty, which may cause issues in downstream logic or user experience.

# Platform-Specific Behavior

This issue was particularly noticeable on Android devices, where losing focus from the NicknameEditor did not trigger nickname generation when the input was left black.

On **iPhones**, this behavior already works as expected.

# Screen recordings

https://github.com/user-attachments/assets/e5df7d67-bf67-4211-b340-477e857fc08d

## Checklist


callebtc:
like it, thank you!

app/src/main/java/com/bitchat/android/ui/ChatHeader.kt
```diff
@@ -117,7 +120,14 @@ fun NicknameEditor(
                     focusManager.clearFocus()
                 }
             ),
-            modifier = Modifier.widthIn(max = 100.dp)
+            modifier = Modifier
+                .widthIn(max = 100.dp)
+                .onFocusChanged { focusState ->
+                    isFocused.value = focusState.isFocused
+                    if (!focusState.isFocused && value.isBlank()) {
+                        onValueChange("anon${Random.nextInt(1000, 9999)}")
```
callebtc: could you refactor this and combine it with the random nickname generated in DataManager.kt :39?

otherwise LGTM!

whflima: I think it would be a good idea to create a utility function for generating random nicknames that can be reused throughout the app. It’s a great way to centralize the random generation logic. Here’s how we can structure it:


```kotlin
// Utils.kt - Utility Function
object NicknameUtils {
    fun generateRandomNickname(): String {
        return "anon${Random.nextInt(1000, 9999)}"
    }
}

// ChatHeader.kt - NicknameEditor Component
.onFocusChanged { focusState ->
    isFocused.value = focusState.isFocused
    if (!focusState.isFocused && value.isBlank()) {
        onValueChange(NicknameUtils.generateRandomNickname())
    }
}

// DataManager.kt - Uses the utility
fun loadNickname(): String {
    val savedNickname = prefs.getString("nickname", null)
    return if (savedNickname != null) {
        savedNickname
    } else {
        val randomNickname = NicknameUtils.generateRandomNickname()
        saveNickname(randomNickname)
        randomNickname
    }
}
```

whflima: I’ve made those changes. Let me know if there’s anything else that needs adjusting.

PR #308, opened 2025-08-24
Fix location detection

mario7421:
This PR fixes a bug where the app was showing an incorrect geohash when a user's location was not yet known to the system. Previously, the app would default to a hardcoded location in San Francisco, which could lead to users connecting to incorrect location-based channels.
Tested on GrapheneOS without Google Play Services.

Changes include:
- Removed the default location in San Francisco
- Added a loading spinner to indicate when location data is being fetched
- Implemented improved location request handling with support for different Android versions

## Checklist
- [x] I have read the contribution guidelines
- [x] There is no corresponding issue
- [x] There are no tests for the changed class, so maybe is not a core feature


callebtc:
awesome thank you, I was wondering what the hell is going on!

PR #329, opened 2025-08-28
feat: `MAX_MESSAGE_LENGTH` and `MAX_DISPLAY_NAME_LENGTH` limits

lollerfirst:
# Description
- Enforces a maximum limit of 280 char per message, 30 char per display name.
- Message input shows a live “xx/280” counter with currently used chars
- Incoming messages that are larger than 280 character are automatically dropped


callebtc:
already implemented max nick length. message drop for long messages is a bit harsh... closing for now.

callebtc:
it would be better to truncate long messages and let the user expand them with a click

PR #334, opened 2025-08-29
Add warning if network unavailable

### LINKED ISSUE #322: Improvement: check for network permissions and guide the user to turn it on
### Checklist

### App version

latest

Graphene OS may have rejected network permissions, geohash chats fail without any error. 

### Why should this be added?

_No response_

_No response_

cargodog:
This PR adds a warning if Nostr relays cannot be reached. This resolves https://github.com/permissionlesstech/bitchat-android/issues/322

There are existing build errors on main, so I am unable to test. Happy to test once those are resolved.

[edit] The original issue asks for a permission check. That is not possible, as there is no API to check the network permission. The best we can do is detect disconnection (ALL relays unreachable) and prompt the user to check their connection settings.

### DIFF
--- a/app/src/main/java/com/bitchat/android/nostr/NostrGeohashService.kt
+++ b/app/src/main/java/com/bitchat/android/nostr/NostrGeohashService.kt
@@ -108,6 +108,18 @@ class NostrGeohashService(
         coroutineScope.launch {
             val nostrRelayManager = NostrRelayManager.getInstance(application)
             
+            // Set up network error callback to display system messages
+            NostrRelayManager.setNetworkErrorCallback { errorMessage ->
+                // Display network error as a system message in chat
+                val systemMessage = BitchatMessage(
+                    sender = "system",
+                    content = errorMessage,
+                    timestamp = Date(),
+                    isRelay = false
+                )
+                messageManager.addMessage(systemMessage)
+            }
+            
             // Connect to relays
             nostrRelayManager.connect()
             
@@ -153,6 +165,9 @@ class NostrGeohashService(
 
                 // 2) Clear all subscription tracking and caches
                 relayManager.clearAllSubscriptions()
+                
+                // Clear network error callback
+                NostrRelayManager.clearNetworkErrorCallback()
 
                 // 3) Clear Nostr identity (npub/private) and geohash identity cache/seed
                 try {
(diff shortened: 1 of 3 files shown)


callebtc:
cool thank you! this is nice but it's not exactly what the issue is about. nostr relays could be unreachable due to many different reasons, but network permissions should be checked in all cases.

cargodog:
@callebtc I added more context in the original issue, but the problem is the network permission cannot be checked, by design.

Network is not a permission on normal Android, and GrapheneOS has added the network toggle at a lower level than Android's permission system intentionally, so that apps cannot distinguish between the permission being denied by the user or internet being unavailable for any other normal reason.

So, unfortunately, I think the only thing we can do is detect a connection failure and prompt the user to ensure the device (and app for Graphene users) has an internet connection.

cargodog:
> nostr relays could be unreachable due to many different reasons

Also, notice this message only appears if _all_ Nostr relays are unreachable. Its totally expected that some relays will be unreachable, but if _all_ relays are unreachable, then the user has a connectivity problem, right?

callebtc:
> > nostr relays could be unreachable due to many different reasons
> 
> Also, notice this message only appears if _all_ Nostr relays are unreachable. Its totally expected that some relays will be unreachable, but if _all_ relays are unreachable, then the user has a connectivity problem, right?

that's maybe a reasonable assumption but it's not what the issue above is about. you might just have lost wifi.

cargodog:
> that's maybe a reasonable assumption but it's not what the issue above is about. you might just have lost wifi.

Yes, and in this case I still think its reasonable to inform the user Bitchat may not behave as expected, because it doesnt have a connection. The entire point is that we cannot know WHY there is no internet, but at least we can tell the user things won't behave well until we have internet 🤷

What behavior would you prefer to see?

callebtc:
it could be many things: 
- network permissions (this should be detectable?)
- tor down
- wifi off
- nostr relays down
- something else

callebtc:
the build seems to fail, could you check?

cargodog:
No, [network permission is not detectable](https://github.com/permissionlesstech/bitchat-android/pull/334#issuecomment-3236917881).

My idea was to detect the case when we cannot reach any relays, and assume this means we have SOME network problem (could be any reason). Then I let the user know that there is a network problem and they should check permissions, connection, etc.

My question remains: if not this, what solution do you want to see?

>  the build seems to fail, could you check?

Something was broken master when I opened the PR. It shouldn't be a problem now that main is fixed. The rebase isn't easy though, because the nostr connection logic has changed significantly. I want to make sure you actually like my proposal, before I go and do that work.

cargodog:
I have reset the branch to where it was before I attempted the rebase, since the rebase ended up being a lot more work than anticipated. This way you can review my _idea_ from the original PR. 

If you like the idea, then I will do the work of rebasing it and fixing the any errors.

Chessing234 (changes requested):
`setNetworkErrorCallback` throws if anything else already registered, which will crash a second chatviewmodel. also the copy is hardcoded english and fires on websocket send fail, not a real connectivity check.

PR #359, opened 2025-08-30
Getting rid of spam

pie-with-jam:
Implemented local spam filtering to hide unwanted messages

1. Blocks messages from known spam bots: Automatically hides messages from users with nicknames containing spam-tester or semicolon
2. Filters suspicious content: Hides messages containing two or more consecutive semicolons (;;)
3. Works across all channels: Filters both incoming and outgoing messages in Nostr geohash channels and Bluetooth mesh

User commands:
/spam add <nickname> - Add a nickname to spam filter
/spam remove <nickname> - Remove a nickname from spam filter
/spamlist - Show all blocked spam bot nicknames

photo_2025-08-31_00-18-28

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
@@ -11,6 +11,7 @@ import com.bitchat.android.protocol.BitchatPacket
 import com.bitchat.android.protocol.MessageType
 import com.bitchat.android.protocol.SpecialRecipients
 import com.bitchat.android.util.toHexString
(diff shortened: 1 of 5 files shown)


nomadram:
Thank you.

callebtc:
a filter like this will work only 10ms

nomadram:
> a filter like this will work only 10ms

Is it possible to solve this by organizing closed rooms in geochats? To connect to which you need to manually press the "connect" button? A kind of room within a room. If not, then how does the spam filter work in BitMap? It removes up to 95% of junk messages. 

For now the bots have gone quiet, but in any case, such a problem should be solved either by a local spam filter or something else.

PR #387, opened 2025-09-07
French (FR) localization

kravhen:
Related to #7 

# Description
I've implemented french localization (and hardcoded string extraction) locally. Always wanted to learn the localization workflow on android and seemed a good idea to give it a try with an app that's still young and hasn't grown too complex yet.

I can turn this into a draft PR for now if it's better and keep an eye on how things evolve and demand for translations grow. 

I see a bunch of other localization/translation issues and PRs so far.
Maybe there  could also can be a tracking/umbrella issue to consolidate different folks' efforts, keep track of progress of different PRs, agree on conventions, avoid duplicate work (e.g. hardcoded string extraction)?

## Checklist

- [n/a] If it is a core feature, I have added automated tests


MutantRabbit767:
this is awesome

Chessing234 (commented):
this one has been overtaken on both halves — worth closing rather than rebasing, i think.

**the fr catalog.** `values-fr/strings.xml` is on main with 350 translated keys. this branch has 209. 69 of those are already on main-fr, and of the 140 it has that main-fr doesn't, 139 are keys that no longer exist in `values/strings.xml` at all — removed from the app over the last year.

that leaves exactly one live key the branch would add: `about_tagline`. and it's stale too — the en source now reads `The sidegroup chat`, while this branch translates the old wording ("messagerie maillée décentralisée avec chiffrement de bout en bout"). so merging it would put a wrong tagline on french builds.

**the string-extraction half.** the 19 kotlin files here replace hardcoded literals with `stringResource` calls. main has already done that extraction — `BluetoothCheckScreen.kt`, for instance, has 13 `stringResource` calls and no bare `Text("…")` left. so that part is done too.

no criticism of the work — it's a september 2025 branch and the repo moved. if you want to pick french back up, main-fr is missing 212 of the 563 keys in `values/strings.xml`, and that list is a clean starting point for a fresh branch.

kravhen:
Agreed.
Closing PR.

PR #275, opened 2025-08-22
Add Translations, App Info Bottom Sheet, Edge-to-Edge UI and Predictive Back Gestures

yet300:
This pr merged with my last pr(https://github.com/permissionlesstech/bitchat-android/pull/236)

# Description
This pull request modernizes the application's user interface and navigation by implementing two key Android features: Edge-to-Edge display and Predictive Back gestures.

As part of this effort, the ChatScreen has been significantly refactored to use the Material 3 Scaffold component. This simplifies the layout structure, improves maintainability, and aligns the codebase with modern Jetpack Compose best practices.

# Key Changes:
1. Edge-to-Edge UI Implementation
   The app now draws behind the system status and navigation bars, creating a more immersive, seamless, and modern look.
   enableEdgeToEdge() is called in MainActivity to activate the feature.
   WindowInsets are now correctly handled throughout the ChatScreen using modifiers like .navigationBarsPadding and .statusBarsPadding to prevent critical UI elements from being obscured by system bars. The keyboard is also handled gracefully with .imePadding.
2. Predictive Back Gesture Support (Android 13+)
   Added support for the predictive back gesture to provide users with a fluid and intuitive navigation experience. Users can now "peek" at the previous screen before fully committing to the back action.
   Enabled the feature globally by setting android:windowEnableOnBackInvokedCallback to true in the app's theme.
   The OnBackPressedCallback in MainActivity was updated to handle the new handleOnBackStarted, handleOnBackProgressed, and handleOnBackCancelled events.
   These gesture events are propagated to the ChatViewModel, which now manages the back gesture state.
   The ChatScreen observes this state and animates its scale and position based on the gesture's progress, providing clear visual feedback to the user.
3. Refactoring ChatScreen to use Scaffold
   The main layout of the ChatScreen has been refactored from a custom Box and Column structure to a standard Material 3 Scaffold.
   Benefits:
   Simplifies layout logic by using standard slots like topBar and bottomBar for the header and input sections.
   Improves the handling of WindowInsets and keyboard visibility, as Scaffold manages content padding automatically.
   Reduces boilerplate code and makes the component hierarchy easier to understand and maintain.

##Supported Languages 
AI - translate with AI
Native - verified by a native speaker

*   **en**: English -- Native
*   **ru**: Русский (Russian) -- AI
*   **uk**: Українська (Ukrainian) -- AI
*   **be**: Беларуская (Belarusian) -- AI
*   **tr**: Türkçe (Turkish) -- AI
[...]

PR #406, opened 2025-09-11
Add App Info Bottom Sheet

yet300:
# Description
add new  AboutSheet design(IOS likes) from https://github.com/permissionlesstech/bitchat-android/pull/275

without division into functions  and strings (for easy merging)

If this is approved, I will create a new one with other improvements in the design of sheets.
Screenshot_20250911_160941
Screenshot_20250911_161002


callebtc:
thank you this looks so much better!

PR #434, opened 2025-09-15
fix: sync own messages

callebtc:
# Description

## Checklist

### DIFF
--- a/app/src/main/java/com/bitchat/android/sync/GossipSyncManager.kt
+++ b/app/src/main/java/com/bitchat/android/sync/GossipSyncManager.kt
@@ -156,8 +156,12 @@ class GossipSyncManager(
             return GCSFilter.contains(sorted, v)
         }
 
+        // Track what we send per original sender peerID (hex)
+        data class SentStats(var announce: Boolean = false, var msgCount: Int = 0)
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/bitchat/android/sync/GossipSyncManager.kt
```diff
@@ -176,6 +181,17 @@ class GossipSyncManager(
                 val toSend = pkt.copy(ttl = 0u)
                 delegate?.sendPacketToPeer(fromPeerID, toSend)
                 Log.d(TAG, "Sent sync message: Type ${toSend.type} to $fromPeerID packet id ${idBytes.toHexString()}")
+                val senderHex = pkt.senderID.joinToString("") { b -> "%02x".format(b) }
```
shubertm, line 184: Better to have proper name for `b` here

shubertm (changes requested):

Chessing234 (commented):
ttl=0 self packets now skip the self-filter in the connection manager, announcement path, and message handling. a neighbor can source a packet with our peer id and ttl 0 and it will be processed. keep the self-drop; if sync needs loopback, use a local path that never comes from gatt.

PR #452, opened 2025-09-30
New Close button like Ios

yet300:
# Description
New close button in sheets


callebtc:
Thank you for helping to equalize the UI from iOS and android.

PR #455, opened 2025-10-04
String.xml with brazilian translate

ferreirabraga:
# Description
Added a new Strings.xml file with Brazilian translate.

## Checklist


Chessing234 (changes requested):
`values-br` is breton, not brazilian portuguese. this locale won't load for pt-BR users — it needs `values-pt-rBR`, and it's also only ~40 keys against current strings.xml.

PR #466, opened 2025-10-12
Code refactor+migrate to data store

xanderlmk:
# Description
- Code refactor for BluetoothMeshService and ChatViewModel.
- Update BluetoothMeshDelegate and create an implementation which overrides BluetoothMeshDelegate rather than just the ChatVM. 
- Updates to MainActivity
- Change Theme saving to data store instead of shared preferences
- Replace Google Gson to native kotlinx Json
- Add a package directory for application

## Why It matters
- As code grows, it's best to separate components for readability and maintainability.
- DataStore allows for automatic flows without the need to have two separate variables compared to just SharedPreferences.
- Kotlin's native kotlinx.serialization library.Serialization is much easier to use and manipulate and is better performing than Google's Gson

## TODO
- Finish migration from SharedPreferences to DataStore
- Completely remove GSON and use JSON

## Suggestions
- Room is a framework that allows pretty good persistence storage and allows for storing for more complex objects and classes. It is just a wrapper for SQLite. 
- More code refactor for better readability.

## Checklist

- [x] Make sure application works as intended using testing devices


app/src/main/java/com/bitchat/android/application/AppContainer.kt
```diff
@@ -0,0 +1,22 @@
...
+import android.app.Application
+import android.content.Context
+import com.bitchat.android.mesh.BluetoothMeshDelegate
+import com.bitchat.android.mesh.BluetoothMeshDelegateImpl
+import com.bitchat.android.mesh.BluetoothMeshService
+import kotlinx.coroutines.CoroutineScope
+
+interface AppContainer {
+    val meshService : BluetoothMeshService
+    val bmd : BluetoothMeshDelegate
```
shubertm, line 12: It may be good if the name was meaningful, like `meshDelegate` or full `bluetoothMeshDelegate`

app/src/main/java/com/bitchat/android/ui/theme/ThemePreference.kt
```diff
@@ -1,42 +1,83 @@
...
+        private const val PREFS_NAME = "bitchat_settings"
     }
 
-    fun set(context: Context, preference: ThemePreference) {
-        val prefs = context.getSharedPreferences(PREFS_NAME, Context.MODE_PRIVATE)
-        prefs.edit().putString(KEY_THEME, preference.name).apply()
-        _themeFlow.value = preference
+    private val prefs = context.getSharedPreferences(PREFS_NAME, Context.MODE_PRIVATE)
+    override val theme = context.dataStore.data.map { preferences ->
+        (preferences[KEY_THEME] ?: prefs.getString("theme_preference", ThemePreference.System.name))
```
shubertm, line 59: Better to have data migrations considering that the app is already distributed to users. And could also help removing usage of `SharedPreferences` completely

shubertm (changes requested):

Chessing234 (commented):
theme migration will always miss. old prefs store ThemePreference.System.name, then you json-decode it and fall back to system. also updateTheme logs e.printStackTrace() (unit) instead of the message. read the legacy string as the enum name, then write datastore once.

PR #476, opened 2025-10-15
Enable file transfer

callebtc:
# Description

## Checklist


Chessing234 (commented):
OpenableColumns.SIZE is not a cap. a 0 or undersized column makes enforceUriPrecheck succeed and you still stream the whole file. always read the fd with a hard byte limit and abort when it exceeds MAX_FILE_SIZE_BYTES.

PR #491, opened 2025-10-20
Message queue: prioritize messages over fragments and file transfers

callebtc:
# Description

## Checklist

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/BluetoothPacketBroadcaster.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/BluetoothPacketBroadcaster.kt
@@ -114,47 +114,85 @@ class BluetoothPacketBroadcaster(
     private val transferJobs = ConcurrentHashMap<String, Job>()
     
     // Replace simple actor FIFO with a priority-aware channel + processor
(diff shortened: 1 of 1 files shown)


Chessing234 (commented):
the concurrency shape is fine, for the record — the PriorityQueue is local to the processor coroutine and producers go through the channel, so swapping the actor out doesn't introduce a race. worth stating in the description, since "replaced actor with PriorityQueue" reads alarming at first glance.

the thing i'd want addressed is starvation. the ordering is strict priority (0 normal / 1 fragment / 2 file transfer) with no aging, so while normal messages keep arriving, fragments and file transfers never drain. that isn't just latency: fragments have a reassembly deadline on the receiving side, so a busy channel can turn a delayed transfer into a *failed* one, and the sender sees no reason why. a chatty geohash channel is exactly the case where someone also tries to send a photo.

the fragment secondary key compounds it — smaller total wins, so a stream of small transfers indefinitely preempts one large one already in flight. aging (bump priority by queue residency) or reserving a slot every N sends would bound both.

smaller thing: computeQueuedBroadcast runs FragmentPayload.decode on every fragment at enqueue, on the broadcast path. probably lost in the noise next to BLE, but it's a decode per packet that didn't exist before.

Issue #495, opened 2025-10-23
Add blaster relay to default relay list


Spl0itable:
### Checklist

- [x] I have used the search function for [**OPEN**](https://github.com/permissionlesstech/bitchat-android/issues) issues to see if someone else has already submitted the same feature request.
- [x] I have **also** used the search function for [**CLOSED**](https://github.com/permissionlesstech/bitchat-android/issues?q=is%3Aissue+is%3Aclosed) issues to see if the feature was already implemented and is just waiting to be released, or if the feature was rejected.
- [x] I will describe the problem with as much detail as possible.
- [x] This request contains only one single feature, **not** a list of multiple (related) features.

### App version

All

### Feature or improvement you want

Within `/app/src/main/java/com/bitchat/android/nostr/NostrRelayManager.kt` please add the blaster relay `wss://sendit.nosflare.com` to both `DEFAULT_RELAYS` and `defaultRelayUrls`

PR #496, opened 2025-10-23
Add new relay URL to NostrRelayManager

Spl0itable:
Add the blaster relay so that outgoing messages in geohash channels for teleported users are sent to bridged bitchat clients. Perhaps, also add as a blaster for all geohash channels that leverage the GPS-related relays with some regex pattern to ensure all locations use the `wss://sendit.nosflare.com` blaster relay when `app/src/main/assets/nostr_relays.csv` is used.

Closes: https://github.com/permissionlesstech/bitchat-android/issues/495

# Description

## Checklist

- [X ] I have read the contribution guidelines: <https://github.com/permissionlesstech/bitchat-android?tab=readme-ov-file#contributing>
- [X ] I have performed a self-review of my code

- [ X] I have mentioned the corresponding issue and the relevant keyword (e.g., "Closes: #xy") in the description (see <https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue>)
- [N/A ] If it is a core feature, I have added automated tests

### DIFF
--- a/app/src/main/java/com/bitchat/android/nostr/NostrRelayManager.kt
+++ b/app/src/main/java/com/bitchat/android/nostr/NostrRelayManager.kt
@@ -37,7 +37,8 @@ class NostrRelayManager private constructor() {
             "wss://relay.damus.io",
             "wss://relay.primal.net",
             "wss://offchain.pub",
-            "wss://nostr21.com"
+            "wss://nostr21.com",
+            "wss://sendit.nosflare.com"
         )
         
         // Exponential backoff configuration (same as iOS)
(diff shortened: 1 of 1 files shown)


callebtc:
Hi, I appreciate your PR but unfortunately we don't want to add custom relays at this point. it would fragment the network if not everyone uses the same relay set.

callebtc:
please keep it coming, contributions very welcome @Spl0itable

Spl0itable:
I saw Jack's original reply too, but I'm not sure the PR was quite understood. The relay would be used to blast people's notes to as many other relays as possible. It would only be used for egress, not ingress. Especially useful for those who "teleport" as they're not even in the real physical location.

PR #504, opened 2025-10-30
untested: Avoid duplicate connections to same peer

callebtc:
# Description

> Address rotation handled. I added peer-aware gating and robust disconnects so only one direct connection per peer remains active.

  What I changed

  - Peer-connected checks in the tracker
      - New APIs:
          - isPeerDirectlyConnected(peerID: String): Boolean
          - getConnectedAddressForPeer(peerID: String): String?
      - These use addressPeerMap and connectedDevices to resolve if a peer already has an active link.
  - Drop duplicate on first ANNOUNCE
      - In BluetoothMeshService.handleAnnounce:
          - On the FIRST ANNOUNCE for a device connection, before binding deviceAddress -> peerID, it checks if that peerID is already connected via another existingAddress.
          - If found and different, it drops the new, redundant connection via connectionManager.disconnectAddress(deviceAddress).
          - Otherwise it binds the mapping and marks that peer as directly connected (and schedules initial sync). This keeps exactly one connection per peer (client or server).
  - Disconnect now handles server connections too
      - BluetoothConnectionManager.disconnectAddress(...):
          - If it’s a client connection, calls gatt.disconnect().
          - If it’s a server connection (no GATT), calls gattServer.cancelConnection(device).
          - Always cleans up tracking afterward.

  Why this is robust

  - Works with rotating MACs: we can’t know the peer at scan/connection time, but as soon as the peer ANNOUNCE arrives over the new connection, we enforce the one-direct-connection-
  per-peer rule and drop the newcomer.
  - Applies across client/server roles. The check uses existing mappings and connected state, and the disconnect path works for either role.
  - Keeps backoff correct: We previously fixed the attempt-tracking bug so disconnecting the duplicate doesn’t reset the retry window.

  Files changed

  - app/src/main/java/com/bitchat/android/mesh/BluetoothConnectionTracker.kt
      - Added isPeerDirectlyConnected and getConnectedAddressForPeer.
  - app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
      - Enforced dropping duplicate connection in ANNOUNCE handling.
  - app/src/main/java/com/bitchat/android/mesh/BluetoothConnectionManager.kt
      - Upgraded disconnectAddress to support server connections.

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/BluetoothConnectionManager.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/BluetoothConnectionManager.kt
@@ -302,7 +302,23 @@ class BluetoothConnectionManager(
      * Public: connect/disconnect helpers for debug UI
      */
     fun connectToAddress(address: String): Boolean = clientManager.connectToAddress(address)
-    fun disconnectAddress(address: String) { connectionTracker.disconnectDevice(address) }
+    fun disconnectAddress(address: String) {
+        connectionScope.launch {
+            try {
+                val dc = connectionTracker.getDeviceConnection(address)
+                if (dc != null) {
+                    if (dc.gatt != null) {
+                        try { dc.gatt.disconnect() } catch (_: Exception) { }
+                    } else {
+                        // Try canceling server-side connection if present
+                        try { serverManager.getGattServer()?.cancelConnection(dc.device) } catch (_: Exception) { }
+                    }
+                }
+            } catch (_: Exception) { }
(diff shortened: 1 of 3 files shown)


Chessing234 (commented):
title says untested, and cleanupDeviceConnection no longer clears pendingConnections so backoff never resets on a real drop. first-announce also drops a second address for the same peer, which kills dual-radio / reconnect. needs tests before merge.

PR #530, opened 2026-01-02
Prevent quit notification from reappearing

callebtc:


callebtc:
Fixed the notification reappearing after the quit action. The service now cancels the periodic notification updater and drops foreground state before the shutdown coordinator runs, and guards future updates while  is in progress.

callebtc:
Fixed the notification reappearing after the quit action. The service now cancels the periodic notification updater and drops foreground state before the shutdown coordinator runs, and guards future updates while ACTION_QUIT is in progress.

callebtc:
Added a signature-protected internal broadcast for force-finish and register the receiver with that permission (plus package-scoped send). This prevents third-party apps from triggering the UI shutdown on API <33.

PR #532, opened 2026-01-02
Fix small-screen header clipping (#519)

Kargatharaakash:
Fixes #519

Summary:
- Reserve header space as `headerHeight + statusBarInset` so content below the header is never overlapped by the system status bar.
- Use `statusBarsPadding()` on the header Surface, and keep `heightIn(min = headerHeight)` to allow accessibility scaling without clipping.
- Constrain title text (main / channel / private headers) to a single centered line inside a padded full-width container so long titles truncate with ellipsis instead of expanding the header and overlapping left/right controls.

Files changed:
- app/src/main/java/com/bitchat/android/ui/ChatScreen.kt
- app/src/main/java/com/bitchat/android/ui/ChatHeader.kt
- gradle.properties

Verification steps:
1. On a small device (e.g., Unihertz Tank Mini) set Developer Options → Smallest width to **300** dp.
2. Open a chat with a long title (for example: `bitchat/@anon4158 #mesh`) and verify that the right-side controls remain visible and the title truncates with an ellipsis.

PR #544, opened 2026-01-04
Fix header alignment on Samsung Flip 7

rebroad:
Add negative offset to right section of MainHeader to correct alignment issue where user count and icons appeared too far to the right on some devices (e.g. Samsung Galaxy Z Flip 7). The offset shifts the entire right section 8dp to the left to match alignment on other devices.


callebtc:
thank you! could you check if #532 already solves your issue? that would be great, I don't have a samsung flip for testing it unfortunately

callebtc:
could you show screenshots?

PR #545, opened 2026-01-04
Wifi aware refactor mesh core

callebtc:
# Description

This PR refactors the mesh networking layer to introduce `MeshCore`, a shared coordinator that abstracts common mesh logic (peer management, security, packet processing, etc.) away from specific transports. This enables multi-transport support (BLE + Wi-Fi Aware) with shared state and seamless routing between them.

### Key Changes:
- **MeshCore Abstraction:** Centralized logic for `PeerManager`, `SecurityManager`, `StoreForwardManager`, and `PacketProcessor` into a reusable `MeshCore` class.
- **Multi-Transport Support:** Both `BluetoothMeshService` and `WifiAwareMeshService` now utilize `MeshCore` to handle protocol-level logic while focusing on their respective hardware transport implementations.
- **Unified Routing:** Introduced `TransportBridgeService` to route packets between different transport layers.
- **Shared Gossip:** Both services share a single `GossipSyncManager` instance via `MeshServiceHolder`.

### Critical Fixes:
- **Memory Leak Fix:** `MainActivity` now properly detaches its `WifiAwareMeshDelegate` in `onPause` to prevent leaking the Activity context through the singleton `WifiAwareController`.
- **Background Persistence & Notifications:** Moved message persistence and background DM notifications for Wi-Fi Aware into the service layer via `MeshCore.Hooks`. This ensures messages are saved to `AppStateStore` and notifications are shown even when the UI is closed.
- **Race Condition Fix:** Guaranteed that `BluetoothMeshService` is initialized before `WifiAwareMeshService` to ensure a consistent source for the shared `GossipSyncManager`.
- **Code Deduplication:** Removed redundant persistence logic from `MainActivity` callbacks.

## Checklist


callebtc (commented):
Great work on abstracting the Mesh logic! The `MeshCore` design is much cleaner and allows for easier addition of new transports.

However, I found a few critical issues regarding lifecycle management and data persistence that need to be addressed before merging.

### Critical Issues

1. **Memory Leak in MainActivity**
   In `MainActivity.kt`, you set the delegate for `WifiAwareMeshService` inside a `lifecycleScope`:
   ```kotlin
   svc.delegate = object : com.bitchat.android.wifiaware.WifiAwareMeshDelegate { ... }
   ```
   However, unlike `meshService.delegate = null` in `onPause()`, you **never nullify** the `WifiAwareMeshService` delegate. Since `WifiAwareController` holds a singleton reference to the service, and the anonymous delegate holds a reference to `MainActivity` (and `ChatViewModel`), this causes the `MainActivity` to leak every time it is destroyed/recreated.

2. **Data Persistence & Background Handling**
   You are handling message persistence for Wi-Fi Aware inside the `MainActivity` delegate:
   ```kotlin
   if (message.isPrivate) { AppStateStore.addPrivateMessage(...) }
   ```
   This is risky because:
   * If the UI is not active (app in background), `MainActivity` might be paused/stopped. If we fix the memory leak by clearing the delegate in `onPause`, **Wi-Fi Aware messages received in the background will be lost** (not saved to DB), because `WifiAwareMeshService` itself doesn't save them.
   * `BluetoothMeshService` handles this correctly by passing a hook to `MeshCore` (`onMessageReceived = { ... AppStateStore ... }`).
   * **Fix:** Move the `AppStateStore` persistence logic into `WifiAwareMeshService` (passing it as a hook to its `MeshCore` instance), exactly like it is done in `BluetoothMeshService`. This ensures messages are saved regardless of UI state and removes code duplication from `MainActivity`.

### Architectural Observations

3. **Split Brain Peer State**
   You have two `MeshCore` instances (BLE and Wi-Fi), each with its own `PeerManager`. The UI (`MeshDelegateHandler`) merges these lists. This works, but it means `BluetoothMeshService` doesn't inherently know about Wi-Fi peers (only via the bridge).
   * Make sure `TransportBridgeService` routing is robust enough that if `BluetoothMeshService` receives a packet for a Wi-Fi peer, it correctly routes it to the bridge (which it seems to do via `dispatchGlobal`).
   * *Suggestion:* Long-term, a single `MeshCore` managing multiple `Transport` instances might be cleaner, but the current approach of bridging two services is a valid "micro-service" style approach within the app.

4. **Shared Gossip Manager Race Condition**
   ```kotlin
   val shared = com.bitchat.android.service.MeshServiceHolder.sharedGossipSyncManager
   meshCore = MeshCore(..., sharedGossipManager = shared, ...)
   ```
   If `WifiAwareMeshService` starts before `BluetoothMeshService` (e.g., race condition in `MainActivity` init or permission grants), `shared` will be null. `WifiAwareMeshService` will create its own. Then `BluetoothMeshService` starts and creates *another* one.
   * **Fix:** Ensure `MeshServiceHolder` initializes the `GossipSyncManager` lazily or explicitly before either service starts, so they definitely share the same instance.

### Code Style / Minor

* **Duplication:** The code in `MainActivity` to save messages is a copy-paste of logic that should be in the service layer. Moving it to `WifiAwareMeshService` hooks fixes this.
* **MeshCore Hooks:** The `Hooks` pattern is flexible, but `WifiAwareMeshService` ignores the `onMessageReceived` hook, leading to the issue above.

Please address these points, especially the memory leak and the background persistence issue.

alltheseas:
Critical Issues

  1. Race Condition in Init Order

  // WifiAwareMeshService relies on BluetoothMeshService being initialized first
  com.bitchat.android.service.MeshServiceHolder.getOrCreate(context)
  val shared = MeshServiceHolder.sharedGossipSyncManager // could be null
  Risk: If init order changes, Wi-Fi Aware gets its own GossipSyncManager instead of sharing.

  2. Socket Cleanup Unclear

  When does peerSockets get cleaned up on disconnect? Need to verify all paths:
  - onLost() callback
  - Peer handle invalidation
  - App backgrounding

  3. No Backpressure on Bridge

  TransportBridgeService.broadcast() calls all transports synchronously. If Wi-Fi is slow, BLE packets queue unbounded.

  4. Duplicate Processing Risk

  Both BLE and Wi-Fi have separate MeshCore instances. Packet arriving via bridge could be processed twice. Need to verify deduplication works across cores.

PR #564, opened 2026-01-06
feat: meshtastic handoff

a1denvalu3:
# Description

Settings screen to scan and connect BLE devices advertising the meshtastic UUID.
Connect to a device, it will be used to send bitchat packets over LoRa. Likewise, we can receive packets through LoRa as if receiving from another peer.

**This implementation is completely untested, and probably doesn't work**


a1denvalu3:
Linking [this issue](https://github.com/permissionlesstech/bitchat/issues/508)

PR #584, opened 2026-01-13
Improve DM Handshake UI and Queue Reliability

callebtc:
## Summary
This PR improves the responsiveness and reliability of private messaging during the Noise session establishment phase.

### Key Changes
- **Animated Handshake Feedback**: Replaced the static "refresh" icon with an animated "..." text indicator in the chat header while a handshake is in progress.
- **Thread-Safe DM Queue**: Refactored `MessageRouter` to use `ConcurrentHashMap` and synchronized lists, ensuring safe concurrent access between the UI and background services.
- **Immediate Delivery**: Implemented instant message flushing upon handshake completion. Messages no longer wait for the 1-second polling cycle.
- **Direct UI Updates**: Added an `onSessionEstablished` callback to the mesh delegate to allow the UI to transition immediately from "handshaking" to "secure" state.

### Verification Results
- Verified that messages sent before a session is established are correctly queued and delivered immediately upon handshake.
- Confirmed the UI remains responsive and the animated indicator works as expected.
- Verified that thread-safety improvements prevent potential race conditions in the `MessageRouter` outbox.

### Reviewer Recommendations
- Focus on the `MessageRouter.kt` changes for thread-safety.
- Check the integration of the new `onSessionEstablished` delegate method in `ChatViewModel` and `MeshDelegateHandler`.

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
@@ -188,6 +188,12 @@ class BluetoothMeshService(private val context: Context) {
         // SecurityManager delegate for key exchange notifications
         securityManager.delegate = object : SecurityManagerDelegate {
             override fun onKeyExchangeCompleted(peerID: String, peerPublicKeyData: ByteArray) {
+                // Immediate notification for UI updates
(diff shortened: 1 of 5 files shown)


Chessing234 (commented):
looks fine, approved @jackjackbits

Issue #595, opened 2026-01-14
BCH-01-013: Autogenerated Screenshots Persist After Panic (Low)


callebtc:
## Vulnerability Description
**Severity: Low**
**Source:** Cure53 Audit BCH-01 (Nov 2025)

Android generates screenshots of the app for the "Recents" / App Switcher screen. These screenshots are stored in the filesystem and may contain sensitive chat content. They are **not cleared** when Panic Mode is used.

### Impact
Forensic analysis could recover the last viewed screen state even after a panic reset.

### Recommendation
1.  Prevent screenshots for the app switcher using `FLAG_SECURE` (though this might block user screenshots too).
2.  Or ensure that Panic Mode attempts to clear system snapshot caches (difficult without root).
3.  Best practice: Use `FLAG_SECURE` for sensitive screens.

callebtc:
**Technical Findings:** `MainActivity.kt` does not currently set `WindowManager.LayoutParams.FLAG_SECURE`.

**Recommendation:** Add `window.addFlags(WindowManager.LayoutParams.FLAG_SECURE)` in `onCreate`. This will prevent the Android system from taking snapshots for the App Switcher (Recents) screen.

PR #600, opened 2026-01-14
security: Implement peer disambiguation and duplicate nickname warnin…

### LINKED ISSUE #589: BCH-01-003: Android Nickname Handling Enables User Impersonation (High)
## Vulnerability Description
**Severity: High**
**Source:** Cure53 Audit BCH-01 (Nov 2025)

The application allows multiple peers to have the same nickname. When sending direct messages via commands (e.g., `/msg`), the application resolves recipients by nickname rather than a unique ID. Furthermore, public chats do not persistently display identity fingerprints.

### Impact
An attacker can change their nickname to match a legitimate user (e.g., "Bob"). If Alice tries to message "Bob", the app might resolve to the attacker's ID. Alice has no visual way to distinguish the real Bob from the imposter in the UI.

### Affected Code
**File:** `app/src/main/java/com/bitchat/android/ui/CommandProcessor.kt`
**File:** `app/src/main/java/com/bitchat/android/mesh/PeerManager.kt`

```kotlin
// CommandProcessor.kt
val peerID = getPeerIDForNickname(targetName, meshService)
// This finds the first match for the nickname, which might be the attacker
```

### Recommendation
1. Display peer fingerprints or unique visual identifiers in all interfaces (chats, composition).
2. Do not resolve peers solely by nickname if ambiguity exists; prompt the user to select the correct peer.
3. Prevent duplicate nicknames locally or warn the user.

moehamade:
Add fingerprint-based nickname disambiguation with "(you)" marker for current user messages

Closes #589

---

## Context

When using `/block`, `/unlock`, `/message`, `/m`, etc., we display a peer selection dialog.

When attempting to set a nickname that already exists, a warning dialog is shown:
- **Abort** → revert to the previously saved peer nickname  
- **Confirm** → allow reuse of the nickname

To prevent nickname impersonation and improve clarity, fingerprint-based disambiguation has been added across the chat UI.

---

## UI Changes – Fingerprint Disambiguation & (you) Marker

| Peer selection (after /command affecting same username) | Duplicate nickname warning |
|---------------|----------------------------|
|  |  |

| Chat view (fingerprint + you) | Members list (fingerprints) |
|------------------------------|-----------------------------|
|  |  |
---

## Checklist


moehamade:
Taking #595 into consideration, Android can capture screenshots of the app and store them on the filesystem. My concern is that by adding the “(You)” marker to the current user’s messages, we may unintentionally introduce a new privacy risk.
In a scenario where the OS takes a screenshot and the device later falls into someone else’s hands, a forensic tool could extract those screenshots and clearly identify which messages were sent by the device owner. This makes user-authored messages easier to attribute compared to messages that look identical to others.
Because of this, it might be safer to remove the “(You)” label altogether or potentially restrict it to debug builds only so that the user’s own messages are visually indistinguishable from other messages in production builds.

callebtc:
I think this may be a bit overkill and cause confusion for most users. It could be enough if we just added the shorthash for all nicknames that collide IMO.

moehamade:
> I think this may be a bit overkill and cause confusion for most users. It could be enough if we just added the shorthash for all nicknames that collide IMO.

The short hash is essential, but how should we handle the case where multiple users share the same nickname and someone runs /message \<nickname>? What should the expected behavior be?
Also, I believe always displaying the short hash is preferable to showing it selectively. Selective display introduces edge cases like when a nickname belongs to someone who sent older messages. Would we then need to retroactively add or remove short hashes from past messages every time a collision occurs or is resolved? That seems brittle and potentially confusing.

Let's discuss @callebtc

callebtc:
> /message <nickname>? What should the expected behavior be?

that's a good point, but I think we already do something like this in nostr too: the command would have to require input of the nickname and shorthash. the autocomplete "click" can input that for the user too.

Chessing234 (commented):
setNickname no longer saves or announces. unless every caller moved to commitNicknameChange, a typed nick is live-only and a restart reverts it. also this duplicates #638 — don't land two nickname-collision stacks.

Issue #590, opened 2026-01-14
BCH-01-001: Private Key Stored in Plaintext (Medium)


callebtc:
## Vulnerability Description
**Severity: Medium**
**Source:** Cure53 Audit BCH-01 (Nov 2025)

The user's Ed25519 signing private key is stored in plaintext within a standard `SharedPreferences` file (`bitchat_crypto.xml`).

### Impact
If a device is rooted or compromised by malware with elevated privileges, the attacker can easily extract the private key. This key is the long-term identity of the user; possessing it allows full impersonation.

### POC
```bash
cat /data/data/com.bitchat.droid/shared_prefs/bitchat_crypto.xml
```

### Recommendation
Store private keys exclusively in `EncryptedSharedPreferences` or use the Android Keystore system. The `SecureIdentityStateManager` already uses encryption; this pattern should be applied to the signing key as well.

callebtc:
**Technical Findings:** `EncryptionService.loadOrCreateEd25519KeyPair` (Line 388) uses standard `SharedPreferences` named `"bitchat_crypto"`. 

**Recommendation:** Migrate this to use `SecureIdentityStateManager` (which uses `EncryptedSharedPreferences`) or wrap the current implementation with `EncryptedSharedPreferences`.

PR #603, opened 2026-01-14
Fix/590-Private-key-stored-as-Plain-text

anupamjhabbl:
### Summary
This PR fixes a security issue where the Ed25519 signing private key was
stored in plaintext SharedPreferences, allowing extraction on rooted or
compromised devices.

### Changes
- Migrated Ed25519 signing key storage to EncryptedSharedPreferences
  backed by Android Keystore (AES256-GCM).
- Added a one-time migration to preserve existing user identities by
  moving legacy keys into encrypted storage on first run.
- Gracefully falls back to generating a new key if migration fails.

### Security Impact
Prevents long-term identity key extraction and impersonation attacks
on rooted or malware-compromised devices.

### Verification
- Verified successful compilation using `./gradlew assembleDebug`
- Verified via manual testing across fresh install and upgrade scenarios.

### Related Issue
Fixes: #590

### DIFF
--- a/app/src/main/java/com/bitchat/android/crypto/EncryptionService.kt
+++ b/app/src/main/java/com/bitchat/android/crypto/EncryptionService.kt
@@ -29,7 +29,8 @@ open class EncryptionService(private val context: Context) {
     companion object {
         private const val TAG = "EncryptionService"
         private const val ED25519_PRIVATE_KEY_PREF = "ed25519_signing_private_key"
-        private const val PREFS_NAME = "bitchat_crypto"
+        private const val OLD_PREFS_NAME = "bitchat_crypto"
+        private const val SECURE_PREFS_NAME = "bitchat_crypto_secure"
     }
     
     // Core Noise encryption service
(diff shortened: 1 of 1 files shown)


callebtc:
### Review & Fix Summary

I have pushed fixes to address the critical issues identified during review:

1.  **Filename Collision (Critical Fix)**
    *   **Issue:** `EncryptedSharedPreferences` was attempting to reuse the existing plaintext preference file (`bitchat_crypto`). This causes crashes because the file format is incompatible (XML vs Encrypted XML).
    *   **Fix:** We now use a distinct filename (`bitchat_crypto_secure`) for the encrypted storage.

2.  **Migration Logic**
    *   **Issue:** The migration logic was attempting to open the same file in conflicting modes.
    *   **Fix:** The migration now explicitly reads from the old plaintext file (`bitchat_crypto`) and writes to the new secure file (`bitchat_crypto_secure`), then clears the old entry.

3.  **Testability / Robolectric Support**
    *   **Issue:** Calling `setUpEncryptedPrefs()` inside `init {}` forced Android Keystore access during object construction. This breaks unit tests (like `SecurityManagerTest`) running in environments without a full Keystore.
    *   **Fix:** Moved `setUpEncryptedPrefs()` inside `initialize()`. This allows test subclasses (e.g., `FakeEncryptionService`) to override `initialize()` and bypass Keystore setup.

PR #617, opened 2026-01-15
Fix: Unicast source-routed packets at origin instead of broadcasting

callebtc:
## Summary
This PR fixes an issue where source-routed packets were being broadcasted (flood-filled) to the entire network by the originating device, instead of being unicast strictly to the first hop defined in the route.

## Changes
- Modified `BluetoothPacketBroadcaster` to check if an outgoing packet originates from the local device (`senderID == myPeerID`) and has a non-empty `route`.
- If a route is present, the broadcaster now identifies the first hop from the route list.
- Attempts to unicast the packet directly to that first hop (checking both server and client connections).
- If the unicast succeeds, the function returns early, preventing the packet from being broadcast to other neighbors.
- If the first hop is not connected, it falls back to the standard broadcast mechanism to ensure delivery attempts continue (though robust source routing implies the first hop should be known).
- Updated `BluetoothConnectionManager` to pass `myPeerID` to the broadcaster for sender verification.

## Impact
- **Reduced Network Congestion:** Source-routed packets (used for specific, targeted delivery) no longer flood the immediate neighborhood unnecessarily.
- **Privacy/Security:** Packets meant for a specific path are not immediately shared with all neighbors, adhering closer to the intent of source routing.
- **Protocol Compliance:** Aligns behavior with standard source routing principles where the path is strictly followed starting from the source.

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/BluetoothConnectionManager.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/BluetoothConnectionManager.kt
@@ -37,7 +37,7 @@ class BluetoothConnectionManager(
     // Component managers
     private val permissionManager = BluetoothPermissionManager(context)
     private val connectionTracker = BluetoothConnectionTracker(connectionScope, powerManager)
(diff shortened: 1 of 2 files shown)


callebtc:
superseded by #619

PR #627, opened 2026-01-15
feat: GATT Client Identity Signaling & Connection Deduplication

callebtc:
## Summary
This PR implements a robust mechanism for GATT clients to signal their stable **Peer ID** to the server immediately after connection. This solves the issue of redundant connections caused by MAC address rotation and allows servers to enforce strict 1-to-1 connection policies per peer.

## Changes
- **Protocol:** Added `IDENTITY_CHARACTERISTIC_UUID` (`...C5E`) to `AppConstants`.
- **Server:** `BluetoothGattServerManager` now exposes the Identity Characteristic and handles write requests. It maps the MAC address to the Peer ID and disconnects the new connection if that Peer ID is already connected via another MAC.
- **Client:** `BluetoothGattClientManager` checks for the Identity Characteristic during service discovery and writes its truncated Peer ID (8 bytes).
- **Tracking:** `BluetoothConnectionTracker` supports updating the Peer ID for an existing device connection (`setDevicePeerID`).
- **Docs:** Added full technical specification at `docs/PEERID_GATT_SERVER_CLIENT_SPEC.md` for cross-platform implementation (iOS).

## Testing
- Verified that clients write their ID upon connection.
- Verified that servers accept the ID and update their internal map.
- Verified that duplicate connections from the same Peer ID (different MAC) are rejected.

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/BluetoothConnectionManager.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/BluetoothConnectionManager.kt
@@ -76,7 +76,7 @@ class BluetoothConnectionManager(
         context, connectionScope, connectionTracker, permissionManager, powerManager, componentDelegate, myPeerID
     )
     private val clientManager = BluetoothGattClientManager(
-        context, connectionScope, connectionTracker, permissionManager, powerManager, componentDelegate
+        context, connectionScope, connectionTracker, permissionManager, powerManager, componentDelegate, myPeerID
     )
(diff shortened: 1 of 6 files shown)


Chessing234 (commented):
identity write puts the raw 8-byte peer id on a public gatt char with WRITE_NO_RESPONSE. anyone hosting the service learns your id before noise. don't advertise identity in clear; bind after handshake.

PR #628, opened 2026-01-16
feat: Request Sync Manager and V2 Packet Updates

### LINKED ISSUE #622: Feature request: Separate treatment of REQUEST_SYNC responses
the request sync protocol is still very brittle. when we ask a peer for packets we haven't seen yet, the only thing that is different about the request sync response (RSR) is that the packets have a TTL=0. They do not explicitly say "I am a RSR and I'm fulfilling REQEUEST_SYNC 1234-abc").

Packet's sent as a RSR need separate treatment:
- message senders might already be offline and have sent a LEAVE
- announcements and messages can be very old (time-based filtering would filter these out)
- RSR might be out of order with the current chat "state". We don't sort incoming messages by packet time but by arrival time, do disincentivize gaming the time stamps. for RSR however, we might want to "slide in" old messages before already seen ones, since they might have occurred minutes or hours ago

related to #525

a1denvalu3:
## Summary
This PR implements the Request Sync Manager to make sync requests "less blind" and improve security by attributing sync responses to specific peers.

### Key Changes
- **BinaryProtocol**: Added `IS_RSR` flag (0x10u) and `isRSR` field to `BitchatPacket`.
- **RequestSyncPacket**: Updated to include `requestId` (for attribution) and future filter fields.
- **RequestSyncManager**: New component to track pending sync requests and validate incoming RSR packets.
- **GossipSyncManager**: Updated to use unicast sync requests registered with `RequestSyncManager` instead of blind broadcasts.
- **SecurityManager**:
    - Enforces timestamp validation (max 2 min skew) for normal packets.
    - Exempts RSR packets from timestamp validation **only if** they are solicited and validated by `RequestSyncManager`.

### Documentation
- Added `docs/REQUEST_SYNC_MANAGER.md` with implementation details.

Closes https://github.com/permissionlesstech/bitchat-android/issues/622

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
@@ -14,6 +14,7 @@ import com.bitchat.android.protocol.MessageType
 import com.bitchat.android.protocol.SpecialRecipients
 import com.bitchat.android.model.RequestSyncPacket
 import com.bitchat.android.sync.GossipSyncManager
+import com.bitchat.android.sync.RequestSyncManager
 import com.bitchat.android.util.toHexString
 import com.bitchat.android.services.VerificationService
 import kotlinx.coroutines.*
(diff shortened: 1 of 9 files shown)


a1denvalu3:
@callebtc Changes are fully retro-compatible.

PR #638, opened 2026-01-17
Nickname Collision Handling and Mention Coloring

### LINKED ISSUE #589: BCH-01-003: Android Nickname Handling Enables User Impersonation (High)

callebtc:
## Summary
This PR implements comprehensive nickname collision handling and identity-aware mention coloring to prevent impersonation and improve chat clarity.

### Key Changes
- **Nickname Disambiguation**: Automatically appends a short ID suffix (last 4 chars of Peer ID) to nicknames in the peer list, chat timeline, and notifications when a collision is detected (e.g., `alice` vs `alice#1234`).
- **Identity-Aware Coloring**: Refactored `ChatUIUtils` to resolve mentions back to Peer IDs. Mentions are now rendered in the actual color of the mentioned user rather than the sender's color.
- **Secure Slash Commands**: Updated `/msg`, `/block`, and `/unblock` to detect ambiguous nicknames and prompt the user to use the unique `#suffix` identifier.
- **Performance Optimization**: Optimized the message rendering pipeline by passing stable nickname snapshots, reducing redundant service calls during scrolling.
- **Self-Collision Support**: The system now detects when another user adopts your nickname and disambiguates their display name while keeping yours clean.
- **Bug Fixes**: Fixed a `ClassCastException` in the sidebar and made the mention autocomplete deterministic.

Fixes https://github.com/permissionlesstech/bitchat-android/issues/589

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
@@ -47,7 +47,7 @@ class BluetoothMeshService(private val context: Context) {
 
     // My peer identification - derived from persisted Noise identity fingerprint (first 16 hex chars)
     val myPeerID: String = encryptionService.getIdentityFingerprint().take(16)
-    private val peerManager = PeerManager()
+    private val peerManager = PeerManager().apply { myPeerID = this@BluetoothMeshService.myPeerID }
     private val fragmentManager = FragmentManager()
     private val securityManager = SecurityManager(encryptionService, myPeerID)
     private val storeForwardManager = StoreForwardManager()
--- a/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
@@ -636,6 +636,10 @@ class BluetoothMeshService(private val context: Context) {
         if (connectionManager.startServices()) {
             isActive = true
             
+            // Add ourselves to peer manager for collision detection
+            val myNickname = try { com.bitchat.android.services.NicknameProvider.getNickname(context, myPeerID) } catch (_: Exception) { myPeerID }
+            peerManager.addOrUpdatePeer(myPeerID, myNickname)
+
             // Start periodic announcements for peer discovery and connectivity
             sendPeriodicBroadcastAnnounce()
             Log.d(TAG, "Started periodic broadcast announcements (every 30 seconds)")
@@ -697,6 +701,14 @@ class BluetoothMeshService(private val context: Context) {
         return reusable
     }
     
+    /**
+     * Update our own nickname in peer manager and announce it
+     */
+    fun updateSelfNickname(newNickname: String) {
+        peerManager.addOrUpdatePeer(myPeerID, newNickname)
+        sendBroadcastAnnounce()
+    }
+
     /**
      * Send public message
      */
@@ -1175,6 +1187,11 @@ class BluetoothMeshService(private val context: Context) {
      * Get peer nicknames
      */
     fun getPeerNicknames(): Map<String, String> = peerManager.getAllPeerNicknames()
+
+    /**
+     * Get disambiguated peer nickname (nickname#suffix if collisions exist)
+     */
+    fun getDisambiguatedNickname(peerID: String): String = peerManager.getDisambiguatedNickname(peerID)
     
     /**
      * Get peer RSSI values  
--- a/app/src/main/java/com/bitchat/android/mesh/PeerManager.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/PeerManager.kt
@@ -89,6 +89,9 @@ class PeerManager {
     // Callback to check if a peer is directly connected (injected by BluetoothMeshService)
     var isPeerDirectlyConnected: ((String) -> Boolean)? = null
 
+    // My own Peer ID (to treat specially in disambiguation and cleanup)
+    var myPeerID: String? = null
+
     // Coroutines
     private val managerScope = CoroutineScope(Dispatchers.IO + SupervisorJob())
     
@@ -305,7 +308,9 @@ class PeerManager {
         val info = peers[peerID] ?: return peerID
         val nick = info.nickname.trim()
         val isAmbiguous = peers.values.count { it.nickname.trim().equals(nick, ignoreCase = true) } > 1
-        return if (isAmbiguous) "$nick#${peerID.takeLast(4)}" else nick
+        
+        // Suffix is appended if ambiguous, UNLESS this is our own Peer ID
+        return if (isAmbiguous && peerID != myPeerID) "$nick#${peerID.takeLast(4)}" else nick
     }
     
     /**
@@ -428,7 +433,9 @@ class PeerManager {
     private fun cleanupStalePeers() {
         val now = System.currentTimeMillis()
         
-        val peersToRemove = peers.filterValues { (now - it.lastSeen) > stalePeerTimeoutMs }
+        val peersToRemove = peers.filterValues { 
+            it.id != myPeerID && (now - it.lastSeen) > stalePeerTimeoutMs 
+        }
             .keys
             .toList()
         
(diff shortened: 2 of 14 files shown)


app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
```diff
@@ -636,6 +636,10 @@ class BluetoothMeshService(private val context: Context) {
         if (connectionManager.startServices()) {
             isActive = true
             
+            // Add ourselves to peer manager for collision detection
+            val myNickname = try { com.bitchat.android.services.NicknameProvider.getNickname(context, myPeerID) } catch (_: Exception) { myPeerID }
+            peerManager.addOrUpdatePeer(myPeerID, myNickname)
```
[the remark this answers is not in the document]
moehamade: This also causes issues like being able to open a chat with yourself, this is best solved at data layer level by filtering getActivePeerIDs() in PeerManager.kt to exclude self. This will solve it for isConnected and all other cases.
`return peers.filterValues { it.isConnected && it.id != myPeerID  }`
```kotlin
    /**
     * Get list of active peer IDs
     */
    fun getActivePeerIDs(): List<String> {
        return peers.filterValues { it.isConnected && it.id != myPeerID  }
            .keys
            .toList()
            .sorted()
    }
```

app/src/main/java/com/bitchat/android/mesh/PeerManager.kt
```diff
@@ -313,6 +301,18 @@ class PeerManager {
         return peers[peerID]?.nickname
     }
     
+    /**
+     * Get disambiguated peer nickname (nickname#suffix if collisions exist)
+     */
+    fun getDisambiguatedNickname(peerID: String): String {
+        val info = peers[peerID] ?: return peerID
+        val nick = info.nickname.trim()
+        val isAmbiguous = peers.values.count { it.nickname.trim().equals(nick, ignoreCase = true) } > 1
```
moehamade, line 310: I think this can be made more efficient,  every time a peer joins/leaves, all visible messages recompose and each one calls:
  `peers.values.count { it.nickname.trim().equals(nick, ignoreCase = true) } > 1`
Consider using a Set to cache ambiguous names instead of calculating this for every message.
Not really a big issue because only 20 messages are shown on the screen anyway, but it's not optimal.

Assuming we have 50 peers:

  Without cache:
  - 20 visible messages × 50 peers = 1,000 string comparisons per peer change

  With cache:
  - Rebuild cache: 50 iterations (once)
  - Lookups: 20 Set checks (O(1) each)
  - Total: 70 operations
  - Savings: 93%

app/src/main/java/com/bitchat/android/ui/ChatScreen.kt
```diff
@@ -167,9 +169,13 @@ fun ChatScreen(viewModel: ChatViewModel) {
...
                 onMessageLongPress = { message ->
                     // Message long press - open user action sheet with message context
-                    // Extract base nickname from message sender (contains all necessary info)
-                    val (baseName, _) = splitSuffix(message.sender)
-                    selectedUserForSheet = baseName
+                    // Use disambiguated name so actions like /block work with suffixes if collisions exist
+                    val disambiguated = if (message.senderPeerID != null && !message.senderPeerID.startsWith("nostr_")) {
+                        viewModel.meshService.getDisambiguatedNickname(message.senderPeerID)
+                    } else {
+                        message.sender
```
moehamade, line 176: Enhancement (Optional): Compose UI should be as dumb as possible, we use viewModels as brains to control what is usually displayed on the screen,  or helper functions for formatting UI stuff. 

I would personally move this to a viewModel
```kotlin
// In ChatViewModel.kt
fun handleMessageLongPress(message: BitchatMessage) {
      val disambiguated = if (message.senderPeerID != null && !message.senderPeerID.startsWith("nostr_")) {
          meshService.getDisambiguatedNickname(message.senderPeerID)
      } else {
          message.sender
      }
        ```
}

// In ChatScreen.kt
onMessageLongPress = viewModel::handleMessageLongPress
```
or use an MVI (model view intent) approach that sends events/ intents from screen UI to the viewModel and proceed this event in the viewModel:
```kotlin

// in ChatScreen
onMessageLongPress = { onAction(MessageLongPressed(message)) }

// in ChatViewmodel
fun onAction(action: ChatActions){
   when(action){
      is MessageLongPressed -> { /** Do some work **/ }
    }
}

File: app/src/main/java/com/bitchat/android/ui/CommandProcessor.kt [hunk 4]
```diff
@@ -268,7 +305,31 @@ class CommandProcessor(
...
+                is PeerResolutionResult.NotFound -> {
+                    val systemMessage = BitchatMessage(
+                        sender = "system",
+                        content = "user '$targetName' not found.",
+                        timestamp = Date(),
+                        isRelay = false
+                    )
+                    messageManager.addMessage(systemMessage)
+                }
+            }
```
moehamade, line 332: Noticed there's a lot of code duplication that can be extracted into a helper function (we're currently doing same checks and creating same messages), and content should be extracted to string resources for translations.

app/src/main/java/com/bitchat/android/ui/ChatUIUtils.kt
```diff
@@ -328,6 +358,34 @@ fun splitSuffix(name: String): Pair<String, String> {
...
+    return nicknames.entries.find { (id, nick) ->
+        if (suffix.isNotEmpty()) {
+            // Match nickname and Peer ID ending
+            nick.equals(base, ignoreCase = true) && id.endsWith(suffix.removePrefix("#"), ignoreCase = true)
+        } else {
+            // Match nickname only
+            nick.equals(displayName, ignoreCase = true)
+        }
+    }?.key
+}
```
moehamade, line 387: I'm not sure if it's correct,I think this is wrong, that's what I got from debugging:
  
  base = "anon6250"                                                                                                                                                                                    
  suffix = "#d116"                                                                                                                                                                                     
  id = "d1169b550ebfb205"                                                                                                                                                                              
  nick = "anon6250"                                                                                                                                                                                    
  displayName = "anon6250#d116"                                                                                                                                                                        
  displayName = "anon6250#d116"

should create:
  `"$nick#${peerID.takeLast(4)}"  // Should be "anon6250#b205"`

it's using:
  `"$nick#${peerID.take(4)}"  // Created "anon6250#d116"`

maybe consider using 
`takeLast` instead of `endsWith` or double checking what we're using on iOS vs Android.

Open from 3 devices (1 iOS and 1 Android at least), and check the 3rd device hashes from Phone 1 and Phone 2,  I think they are displaying different hashes, maybe one platform is taking first 4 and the other is taking last 4?

moehamade (changes requested):
When using /mention , it correctly shows the ambiguous users in the commands, but when using any other command like /m or /block etc.. I think we should do the same in that case.

Critical fixes would be to filter self from active peers and double check displayed hashes if it's similar with iOS.

Medium improvements would be to fix ambiguous names caching with using something like:  
`private val ambiguousNicknamesCache = ConcurrentHashMap.newKeySet<String>()`

Rest are optional improvements but nice to be done.

Chessing234 (commented):
nickname_click still stores the display name, so tapping a disambiguated alice#ab still resolves by string. stash the peer id on the annotation. (moehamade already has the /m /block and self-filter notes — this is the remaining click path.)

PR #642, opened 2026-01-17
feat: giphy gif picker and in-chat rendering

a1denvalu3:


Chessing234 (commented):
this renders every matching giphy url through coil as soon as a recipient opens the conversation. that makes an inbound message trigger a direct request to giphy, exposing the recipient's ip, user agent, and viewing time outside the app's relay/tor path. keep it as a link or require an explicit tap before fetching remote media.

PR #644, opened 2026-01-17
Fix: Active peer notification in background

callebtc:
## Summary
This PR fixes an issue where the "Active Peers" notification (e.g. "bitchatters nearby!") was not triggering when the app was in the background. It also fixes a logic flaw where reconnecting to previously seen peers after a disconnect (0 peers) would not trigger the notification.

## Changes
- **BluetoothMeshService.kt**: Modified `onPeerListUpdated` to explicitly call `serviceNotificationManager.showActiveUserNotification(peerIDs)` when the UI delegate is null (indicating the app is in the background). This ensures the notification logic is executed even when the UI is not active.
- **NotificationManager.kt**: Modified `showActiveUserNotification` to clear `notificationIntervalManager.recentlySeenPeers` when the peer list is empty. This resets the tracking state so that subsequent connections (going from 0 to n peers) are treated as "new" connections, allowing the notification to trigger again.

## Testing
- Verified that notifications now appear when the app is backgrounded and peers connect.
- Verified that disconnecting and reconnecting triggers the notification again.


Chessing234 (commented):
the notification-side changes look sensible — clearing recentlySeenPeers and cancelling when the peer list empties fixes a stale notification that genuinely sticks around.

the part i'd reconsider is inferring background state from `delegate == null` and then *writing* it:

    if (delegate == null) {
        serviceNotificationManager.setAppBackgroundState(true)

a null delegate means no UI delegate is attached, which isn't the same thing as the app being backgrounded — it's also true during startup before the view model attaches, and across a delegate re-attach. the app already tracks this properly: PowerManager derives isAppInBackground from ProcessLifecycleOwner (PowerManager.kt:140/176/180).

that matters because the flag isn't notification-local. MeshDelegateHandler:240-247 reads getAppBackgroundState() to decide `shouldSendReadReceipt = !isAppInBackground && ...`. so a transient null delegate while the app is actually on screen now silently suppresses read receipts until something flips it back — a messaging-visible side effect from a notification fix.

reading the lifecycle state rather than asserting it would avoid both problems.

PR #648, opened 2026-01-18
Migrate identity storage to Tink with one-time legacy migration

yet300:
# Summary:
- Replace EncryptedSharedPreferences with Tink-backed encrypted storage for identity data.
- Add one-time migration from legacy EncryptedSharedPreferences to Tink storage.
- Run migration during Application startup.

# Implementation details:
- Tink AEAD initialized via AndroidKeysetManager with Android Keystore master key.
- Ciphertext stored in standard SharedPreferences; key name used as AAD.
- Migration copies string/string-set values, then deletes legacy prefs.

Notes:
- Keep androidx.security:security-crypto for the migration release; remove after migration is confirmed.

## Checklist


Chessing234 (commented):
moving off `EncryptedSharedPreferences` is worth doing — it's deprecated and the jetpack guidance is to migrate. but the migration path here can permanently destroy a user's identity, and i think that has to be fixed before anything else is discussed.

`DataMigrationManager.handleMigrationFailure`:

```kotlin
when (error) {
    is GeneralSecurityException,
    is IOException,
    is KeyStoreException -> {
        context.deleteSharedPreferences(OLD_PREFS_NAME)
    }
    else -> Unit
}
```

those three are exactly the *recoverable* failures, not the unrecoverable ones. `openOldEncryptedPrefs` throws `GeneralSecurityException` / `KeyStoreException` when the Android Keystore is unavailable — during Direct Boot before first unlock, on a device that's just had a credential change, or on a transient keystore error — and `IOException` for a transient read problem. the response is to delete `bitchat_identity`, which holds `static_private_key`, `signing_private_key` and the whole identity. there is no backup, the noise static key is the user's identity on the mesh, and every favorite relationship keyed on it is orphaned. one transient keystore hiccup at launch and the user is a stranger to everyone they've ever talked to.

it's also self-defeating: the migration flag isn't set on that path, so the next launch retries, finds nothing, and provisions a fresh identity — the delete didn't even buy a working retry.

the case where deleting *is* right — the keystore key is genuinely gone, so the old blob can never be decrypted again — is not distinguishable from a transient keystore error by exception type. if you need a give-up path it should be attempt-counted and explicit ("we tried n times across n launches, the data is undecryptable"), and it should tell the user their identity was lost rather than silently minting a new one. the safe default is to leave the legacy prefs alone and retry, which is what the `!committed` branch already correctly does.

two more things:

**the new store leaks key names.** `EncryptedSharedPreferences` used `AES256_SIV` on *keys* as well as `AES256_GCM` on values. `TinkSecurePreferences` writes into a plain `context.getSharedPreferences(...)`, so the xml now carries readable entries like `static_private_key`, `signing_private_key`, `cached_fingerprint_nicknames`. values stay sealed, but the shape of the identity store becomes visible to anything that can read app-private storage. that's a regression against what's being replaced, and it's worth either hashing the key names or saying explicitly that it's an accepted trade.

**no tests.** this is a one-shot, irreversible transformation of the only thing in the app that can't be regenerated, and the diff has none. at minimum: legacy data present → migrated and readable; migration throws → legacy prefs still there and identity still loads; `commit()` fails → legacy prefs still there; already-migrated → no-op. those are all testable against a fake `SharedPreferences` and they're the four ways this can go wrong in the field.

on the framing — main already declares `tink-android = "1.23.0"` in `libs.versions.toml` and pulls it into a bundle, so the tink dependency is there; the only dependency this branch adds is datastore. worth reconciling, and worth a rebase generally: the branch is CONFLICTING and its base pins `security-crypto = "1.1.0-beta01"` where main is on `1.1.0`.

PR #649, opened 2026-01-18
Add 'Send private message' action to chat user sheet

a1denvalu3:
## Summary
Adds a "Send private message" option to the user action sheet, which appears when long-pressing a user's avatar or message in the chat. This allows users to quickly initiate a Direct Message (DM) without navigating back to the main peer list.

## Changes
- **UI**: Added a "Send private message" row to `ChatUserSheet`. The action uses a distinct **purple** color to separate it from moderation actions (block/slap/hug).
- **Logic**: 
  - Supports **Mesh Chat**: Resolves the peer ID and starts a private chat.
  - Supports **Geohash Chat**: Resolves the user's public key (via nickname or short ID) and initiates a Nostr-based DM.
- **ViewModel**: Added `startGeohashDMByShortId` and `startGeohashDMByNickname` to `ChatViewModel` and `GeohashViewModel`.
- **Repository**: Added `findPubkeyByShortId` to `GeohashRepository` to support looking up users by the 8-char short ID found in `BitchatMessage`.

## Testing
- Verified on local build.
- Checked flow for both Bluetooth mesh peers and Geohash channel participants.

### DIFF
--- a/app/src/main/java/com/bitchat/android/ui/ChatUserSheet.kt
+++ b/app/src/main/java/com/bitchat/android/ui/ChatUserSheet.kt
@@ -112,7 +112,7 @@ fun ChatUserSheet(
                                         // Mesh chat
                                         val peerID = selectedMessage?.senderPeerID ?: viewModel.getPeerIDForNickname(targetNickname)
                                         if (peerID != null) {
-                                            viewModel.startPrivateChat(peerID)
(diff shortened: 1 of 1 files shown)


Michael679089:
Is the message sent directly to your favorite peer or is it still passed peer to peer?

PR #653, opened 2026-01-18
Fix #646: Display shortened pubkey instead of 'anon'

### LINKED ISSUE #646: Bug: Too many "anons".
In the geohash list I'm in there are too many "anons", where if I write they don't reply to me because the message shouldn't be sent and also the little circle appears.

I don't know if this is a bug or not. 

I attach a screenshot. 

@callebtc

a1denvalu3:
## Summary
Fixes #646.

Instead of displaying the generic string "anon" for users who have not yet set a nickname or whose nickname metadata hasn't been received, we now display a shortened version of their public key (e.g., `a1b2c3d4...`).

## Changes
- Updated `GeohashRepository.kt`:
    - Added `formatAnonName` helper.
    - Replaced all fallback instances of "anon" with the shortened pubkey.
    - Updated `refreshGeohashPeople` and display name lookup functions.

## Testing
- Verified that geohash participants without nicknames appear as `abcd1234...` in the list.
- Verified that the user's own name falls back correctly if empty.

### DIFF
--- a/app/src/main/java/com/bitchat/android/nostr/GeohashRepository.kt
+++ b/app/src/main/java/com/bitchat/android/nostr/GeohashRepository.kt
@@ -107,6 +107,10 @@ class GeohashRepository(
         return participants.keys.count { !dataManager.isGeohashUserBlocked(it) }
     }
 
+    private fun formatAnonName(pubkeyHex: String): String {
+        return "${pubkeyHex.take(8)}..."
+    }
+
     fun refreshGeohashPeople() {
         val geohash = currentGeohash
         if (geohash == null) {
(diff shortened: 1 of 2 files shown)


marcorizzo97:
Thank you for taking my request into consideration and fixing that "kind" of bug too.

Michael679089:
I'll check on this later, this pull request is a must have if it's legit.

Chessing234 (commented):
worth knowing that the self branch you changed was already dead in a different way: _nickname is MutableStateFlow<String>("") (ChatState.kt:42), so getNicknameValue() never returns null and the old `?: "anon"` could never fire. switching it to .ifEmpty is what actually fixes it. good change, just a stronger one than the description suggests.

two things it leaves behind though.

the other-peer branch is still `getCachedNickname(pubkeyHex) ?: formatAnonName(...)`, and cacheNickname (GeohashRepository.kt:85) stores whatever it is handed with no empty check — so a peer that announces an empty nickname lands in the map as "", getCachedNickname returns "" rather than null, the elvis doesn't fire, and that person renders with a blank name. that's the same bug you just fixed for yourself, still live for everyone else. .ifEmpty on that branch too, or reject empty in cacheNickname.

second, refreshGeohashPeople isn't the only place this is derived — lines 218, 232 and 233 still do `geoNicknames[lower] ?: "anon"`. after this the same person can show as 1a2b3c4d... in the people list and anon elsewhere, which is a worse end state than consistent "anon" everywhere. #646 is about the label, so those probably want to move together.

PR #663, opened 2026-02-01
Fix/traffic issue optimizations

moehamade:
### Changes

1. **Use 3 relays for all geohash subscriptions** — balances reliability (survives 1 relay down) with bandwidth (limits duplicate event downloads)

2. **Pause presence heartbeat when backgrounded** — `globalPresenceJob` was running 24/7 broadcasting to all channels even when the app was not visible. Now cancelled on background, restarted on foreground.

3. **Split message/presence subscriptions** — Active channel subscription split into kind 20000 (messages, stays open always) and kind 20001 (presence heartbeats, closed on background). Presence events are high-frequency but have no value when the app is not visible.

4. **Subscription lifecycle cleanup** — Presence sub properly cleaned up on channel switch, `currentActiveGeohash` cleared when leaving a location channel, `panicReset` clears relay manager subscriptions to prevent ghost subs on reconnect.

## Checklist

### DIFF
--- a/app/src/main/java/com/bitchat/android/nostr/NostrRelayManager.kt
+++ b/app/src/main/java/com/bitchat/android/nostr/NostrRelayManager.kt
@@ -46,19 +46,8 @@ class NostrRelayManager private constructor() {
         private const val MAX_BACKOFF_INTERVAL = com.bitchat.android.util.AppConstants.Nostr.MAX_BACKOFF_INTERVAL_MS    // 5 minutes
         private const val BACKOFF_MULTIPLIER = com.bitchat.android.util.AppConstants.Nostr.BACKOFF_MULTIPLIER
         private const val MAX_RECONNECT_ATTEMPTS = com.bitchat.android.util.AppConstants.Nostr.MAX_RECONNECT_ATTEMPTS
-        
-        /**
-         * Calculate optimal relay count based on geohash precision
-         * Broad geohashes (1-2 chars) have high activity → need fewer relays
-         * Narrow geohashes (5+ chars) have low activity → need more relays for coverage
-         */
-        fun optimalRelayCount(geohash: String): Int {
-            return when (geohash.length) {
-                1, 2 -> 2  // Broad: continent/country (e.g., "u", "u0", "sy") - high activity
-                3, 4 -> 3  // Medium: province/state - moderate activity
-                else -> 5  // Narrow: city/neighborhood (5+ chars) - low activity, need coverage
-            }
-        }
+
+        fun optimalRelayCount(geohash: String): Int = 3
 
         // Track gift-wraps we initiated for logging
         private val pendingGiftWrapIDs = ConcurrentHashMap.newKeySet<String>()
--- a/app/src/main/java/com/bitchat/android/nostr/NostrRelayManager.kt
+++ b/app/src/main/java/com/bitchat/android/nostr/NostrRelayManager.kt
@@ -168,7 +168,7 @@ class NostrRelayManager private constructor() {
         id: String = generateSubscriptionId(),
         handler: (NostrEvent) -> Unit,
         includeDefaults: Boolean = false,
-    nRelays: Int = 5
+        nRelays: Int = 5
     ): String {
         ensureGeohashRelaysConnected(geohash, nRelays, includeDefaults)
         val relayUrls = getRelaysForGeohash(geohash)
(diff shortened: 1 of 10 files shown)


a1denvalu3:
This is awesome @moehamade. 

Following our discussion there is another potential area of improvement:

1. Split "geohash-*" subscriptions, which currently subscribe to both 20000 and 20001 events into two separate subs listening to 20000 events and 20001 events respectively
2. When backgrounded, close the subscription for kind 20001 events.

app/src/main/java/com/bitchat/android/nostr/NostrRelayManager.kt
```diff
@@ -46,7 +46,9 @@ class NostrRelayManager private constructor() {
         private const val MAX_BACKOFF_INTERVAL = com.bitchat.android.util.AppConstants.Nostr.MAX_BACKOFF_INTERVAL_MS    // 5 minutes
         private const val BACKOFF_MULTIPLIER = com.bitchat.android.util.AppConstants.Nostr.BACKOFF_MULTIPLIER
         private const val MAX_RECONNECT_ATTEMPTS = com.bitchat.android.util.AppConstants.Nostr.MAX_RECONNECT_ATTEMPTS
-        
+
+        fun optimalRelayCount(geohash: String): Int = 3
```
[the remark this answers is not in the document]
moehamade: The flat 3 is intentional. The previous scheme (2/3/5) was counterproductive: 5 relays for narrow geohashes added duplicate downloads with negligible coverage benefit (few messages, nearby reliable relays), while 2 for broad geohashes was fragile (1 relay down = 50% loss, no failover). Relay selection is deterministic per geohash (all users get the same set via closestRelaysForGeohash), so relay count doesn't affect cross-user message delivery. The function still accepts the geohash parameter so it can be made adaptive again in the future if needed.

app/src/main/java/com/bitchat/android/nostr/NostrRelayManager.kt
```diff
@@ -162,7 +168,7 @@ class NostrRelayManager private constructor() {
         id: String = generateSubscriptionId(),
         handler: (NostrEvent) -> Unit,
         includeDefaults: Boolean = false,
-        nRelays: Int = 5
+    nRelays: Int = 5
```
a1denvalu3: indentation

app/src/main/java/com/bitchat/android/ui/GeohashViewModel.kt
```diff
@@ -392,13 +423,29 @@ class GeohashViewModel(
...
     }
 
     override fun onStop(owner: LifecycleOwner) {
-        Log.d(TAG, "🌍 App backgrounded: Pausing sampling for ${activeSamplingGeohashes.size} geohashes")
+        Log.d(TAG, "🌍 App backgrounded: Pausing sampling, presence heartbeat, and presence subscription")
         activeSamplingGeohashes.forEach { subscriptionManager.unsubscribe("sampling-$it") }
+        globalPresenceJob?.cancel()
+        globalPresenceJob = null
+        currentPresenceSubId?.let { subscriptionManager.unsubscribe(it); currentPresenceSubId = null }
     }
```
a1denvalu3: We discussed wanting to pause kind 20001 subscriptions, not the heartbeats coming from our device. I think we want the heartbeat to go on even on background. Here the comment says "presence heartbeat" but anyway I think this is only pausing the samplings and it's just the comment that is misleading.

app/src/main/java/com/bitchat/android/nostr/NostrRelayManager.kt
moehamade: I didn't notice since I didn't edit that part but I'll fix

app/src/main/java/com/bitchat/android/ui/GeohashViewModel.kt
moehamade: I was pausing both, will revert the heartbeat pausing so we only pause Kind 20001 in `currentPresenceSubId?.let { subscriptionManager.unsubscribe(it); currentPresenceSubId = null }`

a1denvalu3 (approved):

a1denvalu3:
ACK @https://github.com/permissionlesstech/bitchat-android/pull/663/commits/4942b71c6e6939a90a430d3b4da9050b35d9f43d . Untested.

Chessing234 (commented):
`LocationChannelsSheet` now starts geohash sampling whenever `availableChannels` or `bookmarks` changes, even while `isPresented` is false. the previous branch ended sampling when the sheet closed; this keeps relay subscriptions running after dismissal and reverses the stated traffic reduction. restore the visibility guard (and call `endGeohashSampling()` when hidden).

PR #668, opened 2026-02-13
build: Use Foojay Toolchains Convention Plugin

moehamade:
This commit adds the `org.gradle.toolchains.foojay-resolver-convention` plugin to the project, which will simplify the management of JDK toolchains by automatically discovering and provisioning them from the Foojay Discovery API.

# Description

## Checklist


Chessing234 (commented):
this collides with the repo's reproducible-build contract, which i don't think the description accounts for.

docs/reproducible-builds.md states the build is pinned to an "exact Temurin JDK release and digest-pinned Linux builder image", and the "Updating dependencies or toolchains" section says to update the exact distribution SHA-256 when the JDK changes. this PR switches provisioning to foojay with `toolchainVendor=JETBRAINS` and `toolchainVersion=21`, resolved through opaque `api.foojay.io/disco/v3.0/ids/<hash>/redirect` URLs. that's a different vendor from the documented one, and a redirect endpoint rather than a digest — so the JDK that compiles the app is no longer pinned by content, and a verifier reproducing a release can't get the same bytes by following the docs.

it also sits outside gradle/verification-metadata.xml. dependency verification is on (verify-metadata=true), but a provisioned toolchain isn't a dependency in that sense, so this is a build input arriving over the network unverified — for an app whose whole value proposition is that you can check what you're running.

the convenience is real and i'm not arguing against toolchain resolution as such. but it needs the docs updated and, ideally, the vendor kept as Temurin with the distribution pinned — otherwise this quietly downgrades a property the project advertises.

PR #677, opened 2026-03-04
Fix foreground service start-timeout crash path

callebtc:
## Summary
- call startForegroundService only when background mode plus Bluetooth/location and notification permissions allow immediate foreground promotion
- promote service to foreground before mesh startup work in onStartCommand to avoid timeout windows
- remove eager mesh holder initialization from onCreate to reduce startup latency before foreground promotion

## Testing
- ./gradlew :app:compileDebugKotlin -q

### DIFF
--- a/app/src/main/java/com/bitchat/android/MainActivity.kt
+++ b/app/src/main/java/com/bitchat/android/MainActivity.kt
@@ -674,6 +674,9 @@ class MainActivity : OrientationAwareActivity() {
                     return@launch
                 }
 
+                // Retry foreground-service startup now that runtime permissions are available.
+                try { com.bitchat.android.service.MeshForegroundService.start(applicationContext) } catch (_: Exception) { }
+
                 // Set up mesh service delegate and start services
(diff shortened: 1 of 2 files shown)


Chessing234 (commented):
adding the bluetooth-permission check to the start gate is right — a FGS started without the permission it declares a type for is exactly the timeout crash.

the retry added in MainActivity is the part i'd push back on. `try { MeshForegroundService.start(...) } catch (_: Exception) { }` swallows everything, so if the start still fails after permissions are granted the user gets no service and no signal, and the next person to debug it has nothing to go on. the branch immediately below already knows how to report failure to the view model; worth at least logging the exception rather than discarding it.

also, start() is now called there and again through startServices() further down — is the second one a no-op if the first succeeded? if both land you get two startForegroundService calls in the same tick, which on some OEM builds is its own source of ANRs.

using the fully-qualified com.bitchat.android.service.MeshForegroundService inline when the file imports elsewhere is a small thing, but it reads like it was patched in rather than written there.

Issue #700, opened 2026-04-28
Too much traffic on mobile data


aand18:
### Checklist

- [x] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/permissionlesstech/bitchat-android/releases/latest).
- [x] I made sure that there are *no existing issues* - [open](https://github.com/permissionlesstech/bitchat-android/issues) or [closed](https://github.com/permissionlesstech/bitchat-android/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] This issue contains only one bug.
- [x] I have read and understood the [contribution guidelines](https://github.com/permissionlesstech/bitchat-android/blob/main/README.md#contributing).

### Affected version

1.7.2

### Steps to reproduce the bug

1. Install Bichat
2. Allow required permissions
3. Bitchat stays active (permanent notification)
4. Settings On:
   - run in background
   - proof of work - default 12 bits
   - Tor Network
5. Join an empty channel (not sure it has any impact)

### Expected behavior

Bitchat should use little or no mobile data when not used. 

If the high data usage is normal behavior for Bitchat, this should be made clear in the readme and maybe as a warning when first time running the app.

### Actual behavior

Found Bitchat used **620MB** of mobile data with basically no use the past weeks. 

For context, WhatsApp which I use occasionally for calls and text stays at 10x less or 60MB (screenshot). GMaps used 200MB in the same time frame and I did quite a bit of travel/car navigation.

### Screenshots/Screen recordings

<img width="1080" height="1713" alt="Image" src="https://github.com/user-attachments/assets/5f3f31a2-1507-42bb-a63c-3e7421ee84d7" />
<img width="1080" height="2367" alt="Image" src="https://github.com/user-attachments/assets/40678a73-8db0-4b7e-899f-5f664a5f7dda" />

### Affected Android/Custom ROM version

Android 15. OneUI 7

### Affected device model

Samsung Galaxy Z Flip6

PR #706, opened 2026-06-08
Pause Nostr geohash firehose in background to cut mobile data Closes: #700

dmcarrington:
The Nostr relay layer ran unbounded in the background: the live geohash channel subscription kept streaming across 5 relays, the presence heartbeat broadcast every ~60s, and a participant-refresh timer polled every 30s. Combined with reconnect/re-subscribe replay on public relays, this could burn gigabytes while idle.

GeohashViewModel now tears down the high-volume subscriptions when the app is backgrounded (onStop) and restores them on foreground (onStart):
- drop the live channel message stream (currentGeohashSubId)
- drop sampling subscriptions
- cancel the presence heartbeat and participant-refresh timer

Gift-wrap DM subscriptions are intentionally kept alive in the background since they are filtered to our pubkey (lightweight) so DMs still arrive.

The selected channel is tracked in activeChannelGeohash so the stream can be rebuilt on foreground without losing the user's selection. The channel subscription logic is extracted into subscribeChannelStream() / subscribeChannelDM() helpers.

# Description

## Checklist

- [x ] I have read the contribution guidelines: <https://github.com/permissionlesstech/bitchat-android?tab=readme-ov-file#contributing>
- [x ] I have performed a self-review of my code

- [ x] I have mentioned the corresponding issue and the relevant keyword (e.g., "Closes: #xy") in the description (see <https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue>)
- [ x] If it is a core feature, I have added automated tests

### DIFF
--- a/app/src/main/java/com/bitchat/android/nostr/NostrFilter.kt
+++ b/app/src/main/java/com/bitchat/android/nostr/NostrFilter.kt
@@ -42,6 +42,33 @@ data class NostrFilter(
                 limit = limit
             )
         }
+
+        /**
+         * Create filter for geohash-scoped chat messages only (kind 20000).
+         * Low-volume; kept subscribed in the background so messages keep arriving.
+         */
+        fun geohashMessages(geohash: String, since: Long? = null, limit: Int = 1000): NostrFilter {
+            return NostrFilter(
+                kinds = listOf(NostrKind.EPHEMERAL_EVENT),
+                since = since?.let { (it / 1000).toInt() },
+                tagFilters = mapOf("g" to listOf(geohash)),
+                limit = limit
+            )
+        }
+
+        /**
+         * Create filter for geohash-scoped presence heartbeats only (kind 20001).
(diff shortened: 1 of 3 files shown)


a1denvalu3:
The problem is that the geohash-$geohash subscription has a filter for both kind 20000 and kind 20001 (message and heartbeat). So the same sub is used for the heartbeats as well. You propose to pause it when backgrounded, but doing so would also pause receiving actual messages from the geohash chat.

When we measured traffic, heartbeats where by far the biggest issue for mobile data.

Maybe we could articulate this PR more, and create a different subcription for sampling the heartbeats of the current geohash, instead of using the same one for the messages. Then we can pause that one?

dmcarrington:
Updated to pause geohash heartbeats (20001) while keeping alive geohash messages (20000) by using different subscriptions

a1denvalu3 (approved):

callebtc:
Thank you, great PR

PR #729, opened 2026-07-12
Bound pre-auth compressed payload expansion

jackjackbits:
## Summary

- reject compressed packets whose declared expansion is empty, negative, or above the uniform 10 MiB protocol ceiling before compressed/output allocation
- reject an empty compressed body before entering the inflater
- require exact declared output length, a complete deflate end marker, and no trailing compressed bytes
- detect under-declared streams even when they exactly fill the output buffer
- keep raw-deflate and legacy zlib-wrapped compatibility under the same exactness checks
- close every native `Inflater` in `finally`
- serialize inflation so concurrent packets cannot multiply the bounded per-packet allocation (memory-budget semaphore pool)
- independently re-check the decompressor result length in `BinaryProtocol`
- catch `Exception` (not `Throwable`) in `decodeCore` so `OutOfMemoryError` is never masked as a parse failure

## Security impact

The previous decoder trusted a four-byte pre-authentication `originalSize` and allocated that amount before packet signature validation. It also accepted a full output buffer when `Inflater.finished()` was false, allowing a larger stream to be silently truncated to an attacker-controlled declaration. This patch closes both the allocation-bomb and truncated-stream acceptance paths with a uniform pre-authentication bound. Outer type `0x22` is not treated as trusted because it is attacker-controlled before inflation/signature verification.

## Compatibility decision (resolved)

**Chosen path: cap new senders at the safe expanded limit with user-visible failure, and explicitly end support for legacy >10 MiB compressed transfers.**

- New senders cap files just below 10 MiB (reserving TLV/encryption-envelope overhead) and `encode()` refuses payloads above the receiver ceiling.
- Picking an oversized file now surfaces a **user-visible error in chat** (previously the send was dropped silently).
- Legacy Android senders (50 MiB UI limit) can still emit highly compressible 10–50 MiB public files that bounded receivers reject; this is an accepted, documented break. A 50 MiB exception based on outer type is unsafe: one attacker-selected frame can allocate 50 MiB, and `BitchatFilePacket.decode` copies content again, taking peak memory above 100 MiB.
- Re-enabling the legacy range later requires authenticated admission before large allocation plus bounded streaming inflation/TLV parsing into temporary storage — documented in `docs/file_transfer.md`.

## Validation

- full debug unit suite: 268 tests, 12 pre-existing skips, 0 failures/errors (includes 11 MiB legacy-fixture policy test)
- `lintDebug`: passed
- raw and zlib exact/under/over/truncated/trailing tests: passed
- negative, zero, empty-body, exact-bound, over-bound tests: passed
- rebased onto current main

### DIFF
--- a/app/src/main/java/com/bitchat/android/protocol/BinaryProtocol.kt
+++ b/app/src/main/java/com/bitchat/android/protocol/BinaryProtocol.kt
@@ -324,21 +324,36 @@ object BinaryProtocol {
         }
     }
     
-    fun decode(data: ByteArray): BitchatPacket? {
+    fun decode(data: ByteArray): BitchatPacket? =
+        decode(data, CompressionUtil::decompress)
+
+    /** Test seam used to prove rejected expansion sizes never reach inflation. */
+    internal fun decodeForTesting(
+        data: ByteArray,
+        decompress: (ByteArray, Int) -> ByteArray?
+    ): BitchatPacket? = decode(data, decompress)
+
+    private fun decode(
+        data: ByteArray,
+        decompress: (ByteArray, Int) -> ByteArray?
+    ): BitchatPacket? {
         // Try decode as-is first (robust when padding wasn't applied) - iOS fix
-        decodeCore(data)?.let { return it }
+        decodeCore(data, decompress)?.let { return it }
         
         // If that fails, try after removing padding
         val unpadded = MessagePadding.unpad(data)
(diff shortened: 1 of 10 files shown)


a1denvalu3:
Follow-up #736 aligns sender/file limits with the 10 MiB receive bound, bounds compressed-input copies before inflation, and avoids redundant fallback inflation. This resolves the immediate rollout gaps; larger files still need chunked application-level transfer.

PR #730, opened 2026-07-12
Bind Noise sessions to claimed peer identities

jackjackbits:
## Summary

- bind every completed Noise XX session to the claimed 16-hex peer ID before transport ciphers become observable
- preserve established sessions while replacement candidates authenticate, then promote only a valid bound candidate
- reject malformed/unbound announcements and known-key replacement before BLE, Wi-Fi, gossip, relay, or last-seen side effects
- require signed leave packets using the already-deployed wire format and clear session state through normal peer removal
- migrate provisional Wi-Fi Aware IDs only after an exact direct Noise completion on the same still-active socket
- restore canonical peer-ID to Nostr indexing only from the authenticated remote Noise key

## Why

Without this binding, a forged handshake or announcement could claim another peer's routing ID, tear down a working session, or mutate transport routing before proving possession of that identity. The Wi-Fi migration path also needs to remain usable for older/provisional discovery IDs without allowing an ambient peer-ID completion or stale socket to authorize a rebind.

## Compatibility and rollout

- no mesh wire values or Noise handshake bytes change
- released Android and iOS clients already derive the same lowercase short ID and already sign leave packets
- response-only handshake APIs remain as compatibility wrappers
- announcement capability bits remain discovery hints only

This PR is intentionally draft and must not ship as a standalone release window. Merge/release it with the stacked private-media change that authenticates peer capabilities and the Ed25519 key inside Noise (`0x21`); announcements alone must never pin downgrade-sensitive state.

## Validation

- focused identity, announcement, leave, packet-side-effect, and exact-link tests: 38 passed
- full unit suite: 144 passed, 12 skipped, 0 failed
- `:app:assembleDebug`: passed
- `:app:lintDebug`: passed with the unchanged repository baseline (424 errors, 277 warnings, 18 hints)
- `git diff --check`


a1denvalu3:
**TL;DR: replayable ANNOUNCE can poison the BLE direct-peer mapping.**

A captured, valid ANNOUNCE remains acceptable for 10 minutes: freshness is checked, but there is no one-time/replay check ([`AnnouncementIdentityValidator.kt`](https://github.com/permissionlesstech/bitchat-android/blob/b87239c/app/src/main/java/com/bitchat/android/mesh/AnnouncementIdentityValidator.kt#L20-L32)). Because the signature and max TTL are preserved by replay, an attacker can retransmit the victim's packet over the attacker's BLE connection.

After announcement acceptance, the code binds the *current BLE device address* (`routed.relayAddress`) to the replayed sender ID with `connectionManager.addressPeerMap[deviceAddress] = pid`, without proving Noise-key possession on that BLE link ([`BluetoothMeshService.kt`](https://github.com/permissionlesstech/bitchat-android/blob/b87239c/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt#L475-L489)). That map is then used to choose the BLE connection for a target peer ([`BluetoothPacketBroadcaster.kt`](https://github.com/permissionlesstech/bitchat-android/blob/b87239c/app/src/main/java/com/bitchat/android/mesh/BluetoothPacketBroadcaster.kt#L188-L203)).

This does **not** break Noise confidentiality or let the attacker complete a handshake as the victim. It can, however, misdirect packets/handshakes, make the victim appear directly connected through the attacker, and disrupt use of an otherwise valid Noise session (availability/routing attack). BLE direct-link promotion should be delayed until Noise authentication completes on that exact BLE connection, analogous to the Wi-Fi ingress-link binding added here.

a1denvalu3:
The BLE-specific requirement should therefore be:

  BLE connection A receives ANNOUNCE claiming peer V
          ↓
  treat V only as a provisional claim
          ↓
  run a fresh Noise handshake associated with connection A
          ↓
  handshake completes and remote static key derives to V
          ↓
  atomically bind BLE connection A → V

PR #734, opened 2026-07-24
feat: scan verification QR from gallery

### LINKED ISSUE #726: Scan qr from images in gallery
### Checklist

### App version

1.7.4

On the scan qr page, I want to be able to scan  a qr that is saved in the device gallery 

### Why should this be added?

 I want to be able to scan  a qr that is saved in the device gallery so that I can transition from closed platforms like WhatsApp to bitchat without needing a second device to display the qr code, reducing friction during transition from closed platforms to bitchat

There can be a button on the bottom right corner that opens the device gallery and allows the user to browse for the image containing the qr code

Neeraj-x0:
## Description

Adds the ability to scan a verification QR code from a gallery image on the verify Scan tab (#726).

Users can pick a saved QR screenshot (e.g. from WhatsApp) without needing a second device to display the code. The image is decoded with the same ML Kit QR path as the live camera scanner, then fed into the existing verification flow.

- Photo button in the bottom-right of the camera scanner opens the gallery
- When camera permission is denied, a **scan from gallery** button remains available
- Shows clear errors when no QR is found or the image cannot be read

Closes #726

## Checklist
- [x] I have mentioned the corresponding issue and the relevant keyword (e.g., ""Closes: #xy"") in the description (see <https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue>)

## Summary
- Gallery image picker on verify Scan tab for QR decode
- Reuses existing `VerificationService.verifyScannedQR` / `beginQRVerification` path
- Works with or without camera permission

## Test plan
- [x] Open verify sheet → Scan tab → tap gallery button → pick an image containing a valid bitchat QR → verification starts
- [x] Pick an image with no QR → see ""no qr code found in image""
- [x] Deny camera permission → still able to use **scan from gallery**
- [x] Camera live scan still works as before

### DIFF
--- a/app/src/main/java/com/bitchat/android/ui/VerificationSheet.kt
+++ b/app/src/main/java/com/bitchat/android/ui/VerificationSheet.kt
@@ -58,6 +58,7 @@ import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
@@ -92,10 +93,14 @@ import com.google.mlkit.vision.common.InputImage
 import com.google.zxing.BarcodeFormat
 import com.google.zxing.common.BitMatrix
 import com.google.zxing.qrcode.QRCodeWriter
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlinx.coroutines.withContext
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.Executors
-
+import kotlin.coroutines.resume
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun VerificationSheet(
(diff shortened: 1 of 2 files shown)


callebtc:
## Android UI verification

Code-level review identifies a visible verification QR gallery flow.

| Before | After |
|---|---|

Expected comparison: the verification sheet without gallery import versus the QR image picker/import and resulting verification state.

Chessing234 (commented):
useful feature, and it lands on a screen with a live problem worth accounting for.

#878 is an open, reproduced crash — two reporters on 2.0.1 — where switching to this exact Scan tab crashes the app instantly. `VerificationSheet` hasn't changed since the redesign in 7025009 (which touched it only cosmetically, 18/18 lines), so it is still present on main.

that matters for where the gallery entry point lives. if the photo button renders inside the permission-granted branch alongside `ScannerView`, then for anyone hitting #878 it's unreachable — the composable that crashes is the one that would have drawn the button. gallery decode is the one path here that needs no CameraX at all, so it could be the working route for those users while #878 is open, but only if it can be reached without mounting the viewfinder. worth making that explicit rather than incidental: the denied-permission branch already proves the decode path stands alone.

i couldn't pin the exact nesting from the diff alone, so treat that as a question rather than a claim — but it's cheap to check and the answer decides whether this helps the people currently stuck.

two smaller things:

it's CONFLICTING against main and will need a rebase before anyone can judge it properly.

the test checkbox is unticked, and there is one genuinely unit-testable piece in here: the URI → bitmap → ML Kit → payload string decode, independent of Compose. the no-QR-found and unreadable-image branches are exactly the ones that will rot silently, and they're the ones you wrote error copy for.

Neeraj-x0:
Will get back to this

PR #735, opened 2026-07-24
feat: restore password-protected channel encryption

Neeraj-x0:
## Description

Restores end-to-end password-protected channel encryption that was stubbed during the protocol rewrite.

Previously, `verifyChannelPassword` always returned true and `sendEncryptedChannelMessage` was a no-op, so `/pass` channels still sent plaintext over the mesh. Encrypt success also ignored the ciphertext and resent plaintext.

This wires PBKDF2-HMAC-SHA256 (100k iterations, channel name as salt) + AES-256-GCM back through send/receive:

- `/pass` derives the channel key and persists a SHA-256 key commitment (not the password)
- Channel messages are sent as `BitchatMessage` binary payloads; encrypted channels use `isEncrypted` + AES-GCM ciphertext
- Receivers decrypt when keyed, otherwise show `[Encrypted message - password required]` and mark the channel password-protected
- Wrong passwords are rejected via commitment or decrypt test against existing ciphertext
- Encrypt failure never falls back to plaintext
- Public mesh (no channel) stays plain UTF-8 for iOS compatibility

## Checklist
- [x] I have mentioned the corresponding issue and the relevant keyword (e.g., ""Closes: #xy"") in the description (see <https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue>)

## Summary
- Restore channel password verify + AES-GCM encrypt/decrypt in `ChannelManager`
- Add `sendBinaryBroadcast` and channel-aware mesh send path
- Decrypt channel binary payloads in `MessageHandler`
- Persist key commitments in `DataManager`
- Unit tests for encrypt/decrypt roundtrip and commitments

## Test plan
- [ ] Create channel, set password with `/pass secret`, send a message — peer without password sees encrypted placeholder
- [ ] Peer joins with correct password — messages decrypt
- [ ] Peer joins with wrong password — rejected
- [ ] Encrypt path never sends plaintext when a channel key is set
- [ ] Public mesh chat (no channel) still works with iOS/plain UTF-8
- [ ] `./gradlew :app:testDebugUnitTest --tests com.bitchat.android.noise.NoiseChannelEncryptionTest`

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
@@ -760,29 +760,58 @@ class BluetoothMeshService(private val context: Context) : TransportBridgeServic
     }
     
     /**
-     * Send public message
+     * Send public message.
+     * Public (no channel): plain UTF-8 for iOS compatibility.
+     * Channel: BitchatMessage binary so receivers get channel metadata.
      */
     fun sendMessage(content: String, mentions: List<String> = emptyList(), channel: String? = null) {
         if (content.isEmpty()) return
-        
-        serviceScope.launch {
-            val packet = BitchatPacket(
-                version = 1u,
-                type = MessageType.MESSAGE.value,
-                senderID = hexStringToByteArray(myPeerID),
-                recipientID = SpecialRecipients.BROADCAST,
-                timestamp = System.currentTimeMillis().toULong(),
-                payload = content.toByteArray(Charsets.UTF_8),
-                signature = null,
-                ttl = MAX_TTL
-            )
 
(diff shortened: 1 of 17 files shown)


callebtc:
## Android UI verification

Code-level review identifies visible channel/encryption interaction behavior.

| Before | After |
|---|---|

Expected comparison: channel controls and password-protected encryption states before and after restoration.

Chessing234 (commented):
joinChannel is now suspend, but there's still a Boolean wrapper that returns channelManager.joinChannel(...) directly. that will not compile. every call site has to go through viewModelScope/coroutineScope, including tests — runChannelManager tests are calling suspend from assertFalse without runBlocking.

Issue #425, opened 2025-09-14
Messages in wrong order


dtonon:
### Checklist

- [x] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/permissionlesstech/bitchat-android/releases/latest).
- [x] I made sure that there are *no existing issues* - [open](https://github.com/permissionlesstech/bitchat-android/issues) or [closed](https://github.com/permissionlesstech/bitchat-android/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] This issue contains only one bug.
- [x] I have read and understood the [contribution guidelines](https://github.com/permissionlesstech/bitchat-android/blob/main/README.md#contributing).

### Affected version

1.3.0

### Steps to reproduce the bug

1. Open the app
2. Join a channel

### Expected behavior

Messages should appear in chronological order.

### Actual behavior

Messages' order is randomly mixed up.

### Additional information

Possibily related to https://github.com/permissionlesstech/bitchat-android/issues/302

Issue #420, opened 2025-09-14
Old messages reappearing randomly.


shortwavesurfer2009:
### Checklist

- [x] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/permissionlesstech/bitchat-android/releases/latest).
- [x] I made sure that there are *no existing issues* - [open](https://github.com/permissionlesstech/bitchat-android/issues) or [closed](https://github.com/permissionlesstech/bitchat-android/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] This issue contains only one bug.
- [x] I have read and understood the [contribution guidelines](https://github.com/permissionlesstech/bitchat-android/blob/main/README.md#contributing).

### Affected version

1.3.0

### Steps to reproduce the bug

Honestly, not sure. I was just chatting for a while and this happened.

### Expected behavior

Old messages should not be appearing again.

### Actual behavior

For some reason old messages just appeared again. Have a look at the timestamps.

### Screenshots/Screen recordings

<img width="1080" height="1091" alt="Image" src="https://github.com/user-attachments/assets/a3f5a9f0-89c7-45d0-8876-3a7cbf726b38" />

### Affected Android/Custom ROM version

LineageOS 22.2

### Affected device model

OnePlus Nord N200

ghost:
true, it's really annoying. From what I see it's messages with higher pow always lower in chat then messages with lower pow

shortwavesurfer2009:
> true, it's really annoying. From what I see it's because messages with higher pow takes longer to receive

Verifying POW takes very little time, so receiving a higher POW message would not take any longer than receiving a lower POW message. Sending a higher POW message takes longer than sending a lower POW message, though.

That's why POW is so great, because it puts computational birding on spam, sending, but it's very quick to verify that you've actually done the work you say you have.

ghost:
@shortwavesurfer2009 I was just stating my observation, it's possible that it's not timing issue, maybe messages processed in sequential order based on pow. I know that it could be wrong, just thought that is worth mentioning for issue investigation.

Updated comment because it was not clear

Issue #302, opened 2025-08-23
Chat in reverse order when coming back to app


cgiangreco:
### Checklist

- [x] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/permissionlesstech/bitchat-android/releases/latest).
- [x] I made sure that there are *no existing issues* - [open](https://github.com/permissionlesstech/bitchat-android/issues) or [closed](https://github.com/permissionlesstech/bitchat-android/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] This issue contains only one bug.
- [x] I have read and understood the [contribution guidelines](https://github.com/permissionlesstech/bitchat-android/blob/main/README.md#contributing).

### Affected version

1.1.0

### Steps to reproduce the bug

1. Chat with people in a channel
2. Exit and close the app
3. Open the app and go back to the same channel
4. The chat history is in reverse order. Last message is shown first, and first message is shown last.

### Expected behavior

When re-entering a channel, the chat history should be loaded in the same order as when you left the app.

### Actual behavior

The chat history is in reverse order. Last message is shown first, and first message is shown last.

### Screenshots/Screen recordings

Original order:
<img width="1080" height="300" alt="Image" src="https://github.com/user-attachments/assets/3fa86605-9967-49e2-9eaf-a86bc7edd717" />

After reopening the app:
<img width="1080" height="428" alt="Image" src="https://github.com/user-attachments/assets/21aafd9e-6271-45e1-95c7-7f6335b16fce" />

### Affected Android/Custom ROM version

Android 15 (Nothing OS 3.2)

### Affected device model

Nothing Phone (2)

VinayHajare:
I think main cause of the bug is in the message display logic. In `MessageComponents.kt`, the **MessagesList** composable uses both `reverseLayout = true` in the `LazyColumn` and also calls `messages.asReversed()`. This results in the message order being reversed twice, so after reopening the app, the chat history appears in reverse order.
Removing the `.asReversed()` call from the `items(messages.asReversed())` line, and just using items(messages) with `reverseLayout = true` can solve this.

RenzMc:
try this fork i fix that https://github.com/RenzMc/RenChat/tree/v1.3.0

PR #738, opened 2026-07-25
fix: order mesh public/channel timeline by source packet timestamp (#525)

### LINKED ISSUE #525: Messages ordered by incoming time instead of packet timestamp
In pre-version 1.6.1, I found a bug where messages are not displayed in chronological order. This means that when you open the app, you see parts of messages from several hours earlier, if not from the previous day. So you see current messages mixed in with very old ones. I don't know if this is a problem with the synchronisation of the messages themselves. 
@callebtc

gunjanjaswal:
Incoming public and channel messages were added to the timeline in receive order. When a peer reconnects and store-forwards or gossip-syncs an older backlog, those hour-old messages ended up at the bottom of the list, interleaved with current ones.

This is parity with iOS, which orders these timelines the same way: `ConversationStore.insert(_:)` does an upper-bound binary search with a `<=` comparison so equal timestamps keep arrival order. `insertByTimestamp` is that algorithm. Android was the one appending in receive order.

Each `BitchatMessage` already carries the source packet timestamp, so the fix inserts each message at its timestamp position instead of appending. It's applied in the `AppStateStore` add paths (which the UI timeline reads from) and the matching `MessageManager` paths, for the **public and channel** timelines.

`insertByTimestamp` keeps a fast path, also matching iOS: a message at or after the tail is appended in O(1), and only an out-of-order message pays for the binary search. In-order arrival is the common case, so the hot path is unchanged.

### On private messages

An earlier version of this PR also ordered DMs by timestamp. While it sat in review, `main` gained `PrivateMessageArrivalOrder` (with `ContactDirectory.canonicalizePrivateChats`), which deliberately orders DMs by **arrival sequence** rather than sender timestamp — its own comment explains why: *"Peer-provided timestamps cannot safely order a conversation because clocks can differ."*

That's a deliberate decision, and it's in direct tension with timestamp-ordering DMs, so I've dropped that part: this PR now covers public and channel only, and leaves DM ordering to that arrival-order path (`canonicalizePrivateChats` would re-sort a timestamp insert straight back anyway). Happy to revisit if the maintainers decide DMs should be timestamp-ordered after all.

### Tests

`MessageOrderingTest` covers out-of-order backlog insertion (public and channel), dedup, stable equal-timestamp ordering, and a mixed sequence that exercises both the fast-path append and the binary-search branch.

Fixes #525. Same underlying problem as #425, #420, and #302.


app/src/main/java/com/bitchat/android/util/MessageOrdering.kt
```diff
@@ -0,0 +1,42 @@
+package com.bitchat.android.util
```
habibgurdov77-star, line 1: Run

habibgurdov77-star (commented):
Fixes #739

gunjanjaswal:
Good flag on clock skew. It is a real consequence of ordering by the source packet timestamp: a peer whose clock runs slow (we accept up to ~10 min skew on announces) can land a live message a few minutes back in the list instead of at the bottom.

I left it as-is because ordering by packet timestamp is what #525 asks for, and it matches the `sortBy { it.timestamp }` that used to live in `MessageRetentionService` before it was commented out, so skew handling felt like a separate change. If you would rather not trust remote clocks that far, an easy guard is to clamp a future-dated timestamp to now (or fall back to receive time past a threshold). Happy to fold that in here if you want it.

callebtc:
## Android UI verification

Code-level review identifies a visible message-timeline ordering behavior change.

| Before | After |
|---|---|

Expected comparison: the same synthetic messages displayed in prior insertion order versus source-packet timestamp order.

gunjanjaswal:
Thanks for running this. Since there's no emulator in the loop here, a quick note on what to look for by hand, because this one won't show up as a pixel diff.

Nothing about the styling or layout changes. The only thing that moves is the order messages land in. To see it: let a peer drop off a channel for a while as messages keep flowing, then bring it back so it store-forwards its backlog. Before this change that hour-old backlog appended to the bottom, interleaved with current messages. After it, each message slots in at its source-packet timestamp, so the backlog drops into its correct chronological place and the tail stays "now". Messages that share a timestamp keep their arrival order (the insertion is a stable binary search), so nothing reshuffles under normal live use.

So there isn't really a before/after screenshot to capture, it's purely timeline ordering, which the insertion logic handles deterministically.

Chessing234 (commented):
this already matches ios, which is worth putting in the body — it makes the change parity rather than a proposed behaviour change.

`ConversationStore.insert(_:)` does the same thing: upper-bound binary search, `<=` comparison so equal timestamps keep arrival order. your `insertByTimestamp` is the same algorithm line for line. i'd cite it; "ios has ordered timelines this way all along and android appends in receive order" is a much easier thing to merge than "here is a new ordering policy".

two differences worth acting on.

**ios applies it to direct conversations too.** `ConversationStore.append(_:to:)` routes every `ConversationID` — including `.direct` — through that same insert. this PR deliberately leaves private chats on append order, so the bug you're fixing stays for DMs on android, and the two platforms would still disagree there. a store-forwarded DM backlog is arguably the *more* likely case, since that's what the courier path exists for.

**ios keeps a fast path.** it only binary-searches when the incoming message is older than the tail:

```swift
if let last = messages.last, message.timestamp < last.timestamp { ...binary search... }
else { messages.append(message) }
```

in-order arrival is the overwhelmingly common case and it stays O(1) there. cheap to add and it keeps the hot path identical to what it is today, which also shrinks the surface of this change.

the dedup-unchanged and stable-equal-timestamp tests are the right things to have pinned.

gunjanjaswal:
All three landed — thanks, these were good calls.

- **Fast path:** `insertByTimestamp` now appends in O(1) when the message is at or after the tail, and only binary-searches when it predates the tail. Hot path is unchanged.
- **DMs:** routed private chats through the same insert, in both `MessageManager` (addPrivateMessage/…NoUnread) and `AppStateStore.addPrivateMessage`. You're right that the store-forwarded DM backlog is the case that most needed it.
- **Framing:** rewrote the description around iOS parity with `ConversationStore.insert` rather than a new policy.

Added tests for the DM backlog ordering and a mixed sequence that hits both the append and binary-search branches. I don't have an Android SDK on this machine so I couldn't run the suite locally, but the JUnit tests will run in CI.

gunjanjaswal:
Following up — rebasing this on `main` turned up a wrinkle worth surfacing, on the DM point specifically.

Since your review, `main` landed `PrivateMessageArrivalOrder` (plus `ContactDirectory.canonicalizePrivateChats`), which deliberately orders DMs by **arrival sequence** instead of sender timestamp. Its own comment makes the case: "Peer-provided timestamps cannot safely order a conversation because clocks can differ." So the project has since made the opposite call for DMs than the iOS-parity one — and `canonicalizePrivateChats` runs on every DM write, so a timestamp insert would just get re-sorted straight back to arrival order.

Rather than fight that, I've pulled the DM ordering back out. The PR now covers public and channel only (where timestamp ordering is uncontested and matches iOS), and leaves DMs to the arrival-order path. If the maintainers would rather DMs were timestamp-ordered after all, that's really a conversation about `PrivateMessageArrivalOrder`, not something to slip in here.

The fast path you suggested stayed in — that one's a clean win regardless. Thanks again for the review; it made the public/channel side better and flagged the DM question that turned out to already have an answer upstream.

gunjanjaswal:
Gentle bump on this one. It orders the mesh public/channel timeline by source packet time (DM ordering left to arrival order, per the earlier thread with @Chessing234), and it's still mergeable. Happy to rebase if it's drifted; just flagging it's ready for a maintainer look whenever there's a moment.

PR #743, opened 2026-07-26
fix: verify Schnorr signatures on incoming geohash Nostr events

### LINKED ISSUE #733: Geohash events (kind 20000/20001) accepted from relay without Schnorr signature verification [High]
- **File:** `app/src/main/java/com/bitchat/android/nostr/GeohashMessageHandler.kt:44`
- **Category:** signature-bypass
- **Severity:** HIGH

**Description:** Incoming geohash chat and presence events from the relay are processed and rendered to the UI without ever calling `NostrEvent.isValidSignature()` or `NostrEvent.isValid()`. Data flow: `NostrRelayManager.handleMessage` (line 653) → `NostrResponse.fromJsonArray` (line 137) constructs `NostrEvent` with wire-supplied `id`, `pubkey`, `sig` → `NostrEventDeduplicator.processEvent` → handler dispatch. The handler `GeohashMessageHandler.onEvent` (line 44) and `NostrClient.handleGeohashMessage` (line 280) never invoke signature verification. The only signature check in the entire nostr package is at `NostrProtocol.kt:77` (for the inner seal of gift wraps). The `PoWPreferenceManager.getCurrentSettings()` check (line 54) is only PoW difficulty, not signature, and defaults to disabled. Per Nostr's threat model, relays are untrusted and clients must verify Schnorr signatures before trusting an event's claimed `pubkey`.

**Exploit scenario:** An attacker who can publish to any connected relay (or compromises a relay) publishes a kind 20000 event with arbitrary `pubkey` (e.g., a victim's known pubkey), arbitrary `n` tag nickname, and arbitrary `content`. The relay delivers it. `GeohashMessageHandler.onEvent` accepts it, caches the nickname, and calls `messageManager.addChannelMessage` with `senderPeerID = "nostr:${pubkey.take(8)}"` and `sender = displayNameForNostrPubkeyUI(pubkey)`. The message renders in the geohash channel as if it came from the impersonated user. 

**Impact:** full pubkey impersonation in geohash channels, fake participant counts, arbitrary content injection, nickname spoofing.

**Recommendation:** Call `event.isValidSignature()` (which recomputes the event ID hash and verifies the Schnorr signature against `event.pubkey`) in `GeohashMessageHandler.onEvent` before any further processing. Similarly add the check in `NostrClient.handleGeohashMessage` and `LocationNotesManager.handleEvent`. Drop events where verification fails.

aharshit123456:
# Description

## Checklist

## Summary
Closes #733.

Geohash chat events (kind `20000`), presence heartbeats (kind `20001`), and
location notes (kind `1`) were processed and rendered to the UI without ever
calling `NostrEvent.isValidSignature()`. Per Nostr's threat model, relays are
untrusted, so a client must verify an event's BIP-340 Schnorr signature
against its claimed `pubkey` before trusting any of its contents.

Without this check, an attacker able to publish to any connected relay could
forge an event claiming a victim's pubkey (with an arbitrary nickname/content)
and have it rendered in a geohash channel as if it came from that victim —
full pubkey impersonation, fake participants, and arbitrary content injection.

## Changes
Added `event.isValidSignature()` checks (the existing BIP-340 Schnorr verifier
already implemented in `NostrEvent`, previously only used for the inner seal of
NIP-17 gift wraps) at all three ingestion points identified in #733:

- `GeohashMessageHandler.onEvent()` — geohash chat + presence events
- `NostrClient.handleGeohashMessage()`
- `LocationNotesManager.handleEvent()` — persistent geohash location notes

Invalid events are logged and dropped before any further processing (dedup,
PoW validation, participant/nickname tracking, or message emission).

Note: I used `isValidSignature()` directly rather than the existing `isValid()`
helper, since `isValid()` also rejects events with empty `content` — which
would incorrectly reject presence heartbeats (kind `20001`), which carry no
content by design.

## Test plan
- Added `GeohashMessageHandlerSignatureTest`, which:
  - constructs a forged event (claims a victim's pubkey, actually signed by an
    attacker's key) and confirms it is **rejected** by `onEvent()` post-fix
    (verified this test fails against the pre-fix code, confirming it
    reproduces the vulnerability)
  - confirms a genuinely-signed event is still accepted and rendered normally
- `./gradlew :app:testDebugUnitTest` — full unit test suite passes, no
  regressions


aharshit123456:
Good catch — fixed in 57a7d3d. Moved the signature check ahead of `dedupe()` so a forged event (bad signature, but a content-derived id that could match a legitimate event) no longer poisons the dedup cache and blocks the genuine copy from being rendered later. Added a regression test reproducing this exact scenario (verified it fails against the old ordering and passes with the fix).

aharshit123456:
@callebtc flagging this one since it addresses a real, currently-exploitable impersonation gap (issue #733). Codex's review feedback on the original diff has already been addressed (see the comment above) and the full test suite passes. Happy to make any further changes if you'd like a different approach.

a1denvalu3:
Move signature verification off the main dispatcher — bitchat-android/
    app/src/main/java/com/bitchat/android/nostr/GeohashMessageHandler.kt:50-50
    When events arrive through GeohashViewModel, this coroutine uses viewModelScope and therefore
    runs on the main dispatcher. isValidSignature() performs synchronous BouncyCastle secp256k1
    scalar multiplications for every event, including the explicitly high-volume presence stream,
    so relay bursts can block UI rendering or trigger an ANR. Perform the cryptographic
    verification on Dispatchers.Default before continuing the state updates on main.

diffusionmam:
great checks, closing the issue #733 as the core issue has been solved

Chessing234 (commented):
right fix and the verifier behind it is sound — `isValidSignature` recomputes the id and compares before checking the signature, so the id-substitution variant (keep a victim's id+sig, swap content) doesn't get through. worth saying explicitly in the body, because that's the part reviewers usually have to go check.

also worth stating: verification sits *before* `dedupe(event.id)` in `GeohashMessageHandler`. that ordering is load-bearing — deduping first would let someone publish a forgery carrying a genuine event's id and have the real one dropped as a duplicate. easy to "optimise" away later without the reason written down.

two things on cost, because rejecting a forgery is now the expensive path and forgeries are the cheap thing to produce.

`NostrClient` verifies at the very top of the handler, ahead of the PoW check and everything else. `GeohashMessageHandler` does it the other way round — kind filter and geohash-tag match first, signature last — which is the order that makes flooding uninteresting. `schnorrVerify` here is pure BigInteger point math with no native backend, so it isn't cheap; putting it after the filters that reject on a string compare costs nothing and drops most junk before the curve math.

related: `GeohashMessageHandler` is constructed twice (`GeohashViewModel` and `NostrBackgroundEventProcessor`), and `NostrClient` now checks the geohash path too, so a single event can be verified up to three times. verifying once at ingress and letting consumers trust the already-validated event would be both cheaper and harder to get wrong than three independent checks that each have to stay correct.

last, minor: the new dispatcher parameter only applies to one of the three sites. `LocationNotesManager` and `NostrClient` still verify inline on whatever thread the relay callback delivers on.

callebtc:
thank you, great PR

Issue #737, opened 2026-07-25
Would it be possible to add multilingual support?


tsjslgy:
### Checklist

- [x] I have used the search function for [**OPEN**](https://github.com/permissionlesstech/bitchat-android/issues) issues to see if someone else has already submitted the same feature request.
- [x] I have **also** used the search function for [**CLOSED**](https://github.com/permissionlesstech/bitchat-android/issues?q=is%3Aissue+is%3Aclosed) issues to see if the feature was already implemented and is just waiting to be released, or if the feature was rejected.
- [x] I will describe the problem with as much detail as possible.
- [x] This request contains only one single feature, **not** a list of multiple (related) features.

### App version

1

### Feature or improvement you want

Users from different countries have different usage scenarios. To make the software more accessible and easier for general users to get started, especially in some regions with special circumstances, would it be worth preparing versions in multiple languages?

aharshit123456:
Hi @tsjslgy I'll Try to take up this issue. 
I noticed that [`app/src/main/res/values/strings.xml`](https://github.com/permissionlesstech/bitchat-android/blob/main/app/src/main/res/values/strings.xml) already supports 33+ languages. with 25 of them having 80% support and rest sitting at 10-12%. I can contribute to the rest and refill them up (Hebrew, Polish, Chinese-Simplified, Chinese-Traditional (12%); Malay, Tamil, Ukrainian) to move the needle a bit. 

But the best case would be to add Weblate/Crowdin which is why I would urge the maintainers of this repo to facilitate me in getting it up on the platform's free tier. It'll make it muchh easier to crowd source translations for as many languages as possible.

Additionally, just adding it here in case you happen to be from the region of supported languages, you can simply switch to that language via changing the language in your Android settings.

I'll attach a PR (that would need review, to backfill for the languages I mentioned and will try to add some of my own too).
Thanks

aharshit123456:
Following up on my earlier comment:

**PR #744 is up** — backfills Hebrew, Polish, Simplified Chinese, Traditional Chinese, Malay, Tamil, and Ukrainian from ~10-12% to 100% (395/395 keys each), plus a few pre-existing bugs found along the way (a Simplified/Traditional script mixup in `values-zh-rTW`, a mistranslated `version_prefix` in `values-pl`, a duplicate-key issue in `values-ms`). Everything in it is machine-translated and flagged as such in-file, pending native-speaker review — it moves the floor up, it doesn't replace a real review pass.

[...]

PR #744, opened 2026-07-26
i18n: backfill 7 severely incomplete locale translations (he, pl, zh-rCN, zh-rTW, ms, ta, uk)

aharshit123456:
# Description

## Checklist
- [x] I have read the contribution guidelines: <https://github.com/permissionlesstech/bitchat-android?tab=readme-ov-file#contributing>
- [x] I have performed a self-review of my code
- [x] I have mentioned the corresponding issue and the relevant keyword (e.g., "Closes: #xy") in the description (see <https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue>) — note: this uses "Relates to #737" rather than "Closes", since it only addresses the translation-completeness part of that issue; the in-app language switcher and translation-pipeline pieces discussed there remain open
- [ ] If it is a core feature, I have added automated tests — not applicable; this is a resource-only translation change, verified via the structural checks below instead of unit tests

## Summary
Relates to #737 (multilingual support request).

The repo already has translation files for 33 languages, but 7 of them were only ~10-12% complete (38-48 of 395 string keys), silently falling back to English for nearly the entire app: **Hebrew, Polish, Simplified Chinese, Traditional Chinese, Malay, Tamil, Ukrainian**.

This PR backfills all 7 to 100% (395/395 keys each), including the two app-wide plurals (`notification_and_more`, `people_count`) with locale-correct CLDR plural categories (e.g. Polish/Ukrainian use one/few/many/other; Chinese/Malay use only other; Hebrew/Tamil use one/other).

## Additional fixes found along the way
- **`values-zh-rTW` script bug**: ~27 keys in the `verify_*`/`fingerprint_*` block (QR/fingerprint verification dialog) were Simplified Chinese characters pasted into the Traditional Chinese file (e.g. `验证` instead of `驗證`). Corrected to proper Traditional Chinese.
- **Leftover untranslated block**: a ~30-key `verify_*`/`fingerprint_*` block existed in raw English in `pl`, `ms`, `ta`, `uk` from before this change (it technically "existed" so wasn't part of the missing-key diff) — translated now.
- **`values-pl` `version_prefix`**: was mistranslated as `"w%1$s"` (translating the English word "version" literally) instead of preserving the literal version-string prefix `"v%1$s"` (e.g. "v1.7.5") — fixed.
- **`values-ms` duplicate key**: a stale untranslated `<string name="notification_and_more">` coexisted with the new `<plurals name="notification_and_more">` — removed the stale duplicate.

[...]

PR #748, opened 2026-07-26
i18n: declare android:localeConfig with all 33 supported locales

aharshit123456:
# Description

## Checklist
- [x] I have mentioned the corresponding issue and the relevant keyword (e.g., "Closes: #xy") in the description (see <https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue>) — uses "Relates to #737" since this addresses only the language-picker piece of that issue, not the translation-completeness or Weblate-pipeline parts
- [ ] If it is a core feature, I have added automated tests — not applicable; this is a manifest/resource declaration with no runtime logic to unit test, verified via build/lint/APK-assembly checks below instead

## Summary
Relates to #737.

Android 13+ (API 33) ships a built-in per-app language picker (Settings → Apps → bitchat → Language). Without an `android:localeConfig` declaration, this picker has no way to know which languages bitchat actually supports, so it falls back to listing every system language instead of just the ones bitchat ships translations for — a much noisier, less useful list for users trying to switch bitchat's language independently of their device's overall language.

## Changes
- Added `app/src/main/res/xml/locales_config.xml`, listing the 33 languages that currently have a `values-<locale>/strings.xml` folder, plus `en` for the default/fallback `values/strings.xml`.
- Referenced it via `android:localeConfig="@xml/locales_config"` on the `<application>` tag in `AndroidManifest.xml`.
- Locale names use BCP-47 tags where they differ from Android's resource-folder naming convention: `pa-rPK` → `pa-PK`, `zh-rCN` → `zh-CN`, `zh-rTW` → `zh-TW`.

No behavior change below API 33 — `minSdk` is 26, and the attribute is simply inert/ignored on older devices, which continue to follow the whole device's system language exactly as before this change.

## Test plan
- [x] `locales_config.xml` parses as well-formed XML
- [x] `./gradlew :app:processDebugResources` — clean resource/manifest merge
- [x] Confirmed `android:localeConfig="@xml/locales_config"` survives manifest merging by inspecting the packaged manifest output
- [x] `./gradlew :app:assembleDebug` — full debug APK builds successfully with the new resource packaged
- [x] `./gradlew lintDebug` — the only lint signal attributable to this change is an expected, non-fatal `UnusedAttribute` info-level note that `localeConfig` has no effect below API 33 (by design, since `minSdk` 26 is below 33)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -75,6 +75,7 @@
         android:icon="@mipmap/ic_launcher"
         android:roundIcon="@mipmap/ic_launcher_round"
         android:label="@string/app_name"
+        android:localeConfig="@xml/locales_config"
         android:supportsRtl="true"
         android:theme="@style/Theme.BitchatAndroid"
         tools:targetApi="31">
--- a/app/src/main/res/xml/locales_config.xml
+++ b/app/src/main/res/xml/locales_config.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<locale-config xmlns:android="http://schemas.android.com/apk/res/android">
+    <locale android:name="en"/>
+    <locale android:name="ar"/>
+    <locale android:name="bn"/>
+    <locale android:name="de"/>
+    <locale android:name="es"/>
+    <locale android:name="fa"/>
+    <locale android:name="fil"/>
+    <locale android:name="fr"/>
+    <locale android:name="he"/>
+    <locale android:name="hi"/>
+    <locale android:name="id"/>
+    <locale android:name="it"/>
+    <locale android:name="ja"/>
+    <locale android:name="ka"/>
+    <locale android:name="ko"/>
+    <locale android:name="mg"/>
+    <locale android:name="ms"/>
+    <locale android:name="ne"/>
+    <locale android:name="nl"/>
+    <locale android:name="pa-PK"/>
+    <locale android:name="pl"/>
+    <locale android:name="pt"/>
+    <locale android:name="pt-BR"/>
+    <locale android:name="ru"/>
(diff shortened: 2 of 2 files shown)


aharshit123456:
@callebtc small, low-risk addition — declares the 33 languages bitchat currently ships translations for via `android:localeConfig`, so Android 13+'s per-app language picker shows that list instead of every system language. No effect below API 33 (minSdk 26). Should be independent of #744 and easy to review on its own.

Chessing234 (commented):
I verified the list and it's correct as of this commit — 33 locale dirs under `app/src/main/res` (34 `values-*` minus `values-night`), and the three BCP-47 rewrites (`pa-rPK` → `pa-PK`, `zh-rCN` → `zh-CN`, `zh-rTW` → `zh-TW`) are right.

The problem is that it's correct *as of this commit*. A hand-maintained copy of the locale list drifts the moment anyone adds a `values-xx/` folder, and nothing in the build catches it — the app silently keeps advertising the old set to the system language picker, which is the exact failure this PR is fixing.

This repo is on AGP 9.3.1 (`gradle/libs.versions.toml:3`), which supports generating the file from the resource folders at build time:

```kotlin
android {
    androidResources {
        generateLocaleConfig = true
    }
}
```

with the default locale declared in `app/src/main/res/resources.properties` (`unqualifiedResLocale=en`). AGP then derives the locale set and injects the manifest attribute, so the list can't go stale and the manual `android:localeConfig` line and `locales_config.xml` both go away.

Could you try that instead? If it doesn't work for some reason here I'd rather know why than guess — and if the manual file has to stay, it needs a test asserting it matches the `values-*` folders, otherwise the drift is guaranteed.

Unrelated: please drop the `🤖 Generated with Claude Code` trailer from the description.

PR #750, opened 2026-07-26
chore(deps): upgrade to AGP 9.3.1, Gradle 9.6.1, Kotlin 2.4.10, SDK 37

moehamade:
Brings the toolchain and every dependency to latest stable.

## Toolchain

| | From | To |
|---|---|---|
| AGP | 8.10.1 | 9.3.1 |
| Gradle | 8.13 | 9.6.1 |
| Kotlin | 2.2.0 | 2.4.10 |
| compileSdk / targetSdk | 35 | 37 |
| Java | 8 | 11 |

API 37 (Android 17) is stable, not preview.

## AGP 9 migration

- Dropped `org.jetbrains.kotlin.android` — AGP 9 provides built-in Kotlin and the plugin is incompatible with the new DSL
- `kotlinOptions.jvmTarget` → `kotlin.compilerOptions` (the String setter is a hard error in Kotlin 2.4)
- Dropped `android.enableJetifier` (deprecated, removed in AGP 10, no support-library deps remain)

`kotlin-parcelize` and the Compose plugin needed no changes. KGP resolves to 2.4.10 transitively, so no `buildscript` pin. No legacy variant API, no custom `BuildConfig` fields, and none of the removed AGP 9 properties were in use.

## Libraries

Compose BOM `2025.06.01` → `2026.06.01` · okhttp `4.12.0` → `5.4.0` · Tink `1.10.0` → `1.23.0` · CameraX `1.5.2` → `1.6.1` · coroutines `1.10.2` → `1.11.0` · core-ktx `1.16.0` → `1.19.0` · lifecycle `2.9.1` → `2.11.0` · activity-compose `1.10.1` → `1.13.0` · navigation-compose `2.9.1` → `2.9.8` · gson `2.13.1` → `2.14.0` · gms-location `21.3.0` → `21.4.0` · security-crypto `1.1.0-beta01` → `1.1.0`

Notable, beyond version numbers:

- **BouncyCastle `1.70` → `1.85`**, switching `bcprov-jdk15on` → `bcprov-jdk18on`. The `jdk15on` line is abandoned; same packages, so no source changes.
- **`mockito-inline` → `mockito-core 5.23.0`.** The former is deprecated — the inline mock-maker has been the default since Mockito 5.
- **`coroutines-test` was pinned at `1.6`** while the runtime was `1.10.2`. Now shares the `kotlinx-coroutines` ref so it can't drift again.
- **`exifinterface`** moved from a hardcoded `1.3.7` into the version catalog at `1.4.2`.

## Wi-Fi Aware permissions

Automated review flagged that `targetSdk 37` brings mandatory local network protection, while `WifiAwareMeshService` reaches peers over link-local IPv6 sockets. `ACCESS_LOCAL_NETWORK` is now declared and requested.

Two related bugs surfaced while verifying this on hardware:

- **Enabling Wi-Fi Aware from Debug Settings never requested its permissions.** The permission flow is gated behind the toggle already being on, and it defaults to off — so onboarding never asked and the toggle didn't either. The controller logged `Missing NEARBY_WIFI_DEVICES permission` on a 5s retry loop forever. Reproduced on two devices; both needed `adb shell pm grant` before the transport would start at all. The toggle and Start chip now request the permissions first.
- **Both permissions needed version gating.** `NEARBY_WIFI_DEVICES` only exists from API 33 and `ACCESS_LOCAL_NETWORK` from API 37, against minSdk 26.

Measured on Android 17: the two are granted independently (`pm grant` of one does not cascade), but share the `NEARBY_DEVICES` group, so requesting them together produces a single prompt. A denied `ACCESS_LOCAL_NETWORK` does not block enabling — the transport starts fine without it and its necessity for link-local sockets is still unproven.

## Verification

- `compileDebugKotlin` — clean, zero warnings
- `testDebugUnitTest` — 124 tests, 0 failures
- `bundleRelease` — builds with R8 and resource shrinking
- `gradlew help`, `build --dry-run` — pass

The 6 R8 `error occurred when parsing kotlin metadata` warnings present under AGP 8.13.2 are gone under 9.3.1.

## Device testing

Pixel 9a on **Android 17 (API 37)** running this branch, so the target behaviours were genuinely exercised:

- Bluetooth mesh — discovery and connection ✅
- Nostr relay chat — end-to-end, also covers okhttp 4 → 5 ✅
- Wi-Fi Aware — permission flow and discovery ✅ (single prompt, both granted)

**Wi-Fi Aware data path does not establish** between the Pixel (Android 17) and a Samsung SM-A366E (Android 16). `ConnectivityManager.requestNetwork` returns `onUnavailable` on both client and server roles, including after role reversal. This happens *before* `connectAwareClientSocket` runs, so no link-local socket is attempted and no `SecurityException` or `EPERM` appears on either device — it is not permission-related. Whether it predates this branch is unconfirmed. Wi-Fi Aware is debug-gated and off by default.

Not yet exercised: Tor via Arti, background voice-note playback, foreground-service survival.

## targetSdk 36/37 behaviour audit

Checked against the source: edge-to-edge and predictive back are already handled correctly; no `BluetoothSocket`/RFCOMM; no `scheduleAtFixedRate`; Safer Intents is opt-in; the reflection in `ChatViewModel` touches instance rather than `static final` fields. On large screens (sw≥600dp) orientation constraints are now ignored — no `screenOrientation` locks exist, so nothing breaks.

## Known hold-back

Robolectric stays at **4.15**. 4.16+ breaks `EncryptionServiceTest` with `AndroidKeyStore not found` — bisected away from `security-crypto`, and not SDK-level related (`@Config(sdk=[35])` with 4.16.1 still fails). Robolectric has never genuinely supported `AndroidKeyStore` ([robolectric#1517](https://github.com/robolectric/robolectric/issues/1517), open since 2015), so the 4.15 pass is incidental. Unpinning needs a storage seam in `EncryptionService`, deferred rather than weakening a test that guards panic-mode key rotation.

## Follow-ups (not in this PR)

- `nordic-ble` is unused — no imports, no ProGuard references — yet resolves onto the release runtime classpath.
- Stale `tor-android-binary` catalog entry, unreferenced and unresolvable. Unrelated to how Tor actually works here (a self-compiled Arti build in `app/src/main/jniLibs/` with the JNI bridge vendored in-tree). Nearby comments in `app/build.gradle.kts` are also stale.
- `AndroidManifest.xml` declares `.service.NotificationPermissionChangedReceiver`, but no such class exists anywhere in the source.
- `lint-baseline.xml` was generated `by="lint 8.2.0"`, four AGP majors ago.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,6 +1,7 @@
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+
 plugins {
     alias(libs.plugins.android.application)
-    alias(libs.plugins.kotlin.android)
     alias(libs.plugins.kotlin.parcelize)
     alias(libs.plugins.kotlin.compose)
 }
@@ -65,11 +66,8 @@ android {
     }
 
     compileOptions {
-        sourceCompatibility = JavaVersion.VERSION_1_8
-        targetCompatibility = JavaVersion.VERSION_1_8
-    }
-    kotlinOptions {
-        jvmTarget = "1.8"
+        sourceCompatibility = JavaVersion.VERSION_11
+        targetCompatibility = JavaVersion.VERSION_11
     }
     buildFeatures {
         compose = true
@@ -83,6 +81,12 @@ android {
         baseline = file("lint-baseline.xml")
         abortOnError = false
         checkReleaseBuilds = false
+    }
+}
+
+kotlin {
+    compilerOptions {
+        jvmTarget.set(JvmTarget.JVM_11)
     }
 }
 
@@ -143,7 +147,7 @@ dependencies {
     implementation(libs.androidx.security.crypto)
     
     // EXIF orientation handling for images
-    implementation("androidx.exifinterface:exifinterface:1.3.7")
+    implementation(libs.androidx.exifinterface)
     
     // Testing
     testImplementation(libs.bundles.testing)
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -31,8 +31,9 @@
       Android 17 (API 37) makes local network protection mandatory for apps targeting it.
       The Wi‑Fi Aware transport reaches peers over link-local IPv6 TCP sockets
       (WifiAwareMeshService.connectAwareClientSocket), so declare this defensively.
-      It belongs to the NEARBY_DEVICES group, but is granted independently of
-      NEARBY_WIFI_DEVICES, so it must be requested explicitly (see PermissionManager).
+      It is granted independently of NEARBY_WIFI_DEVICES so must be requested explicitly
+      (see PermissionManager), but shares the NEARBY_DEVICES group with it, so the two
+      are requested together in a single "Nearby devices" prompt.
     -->
     <uses-permission android:name="android.permission.ACCESS_LOCAL_NETWORK" />
     <!-- Signature permission for internal UI shutdown broadcasts -->
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -27,6 +27,14 @@
     <uses-permission android:name="android.permission.CHANGE_WIFI_STATE" />
     <!-- Android 13+ runtime permission for Wi‑Fi operations (including Aware) -->
     <uses-permission android:name="android.permission.NEARBY_WIFI_DEVICES" />
+    <!--
+      Android 17 (API 37) makes local network protection mandatory for apps targeting it.
+      The Wi‑Fi Aware transport reaches peers over link-local IPv6 TCP sockets
+      (WifiAwareMeshService.connectAwareClientSocket), so declare this defensively.
+      It belongs to the NEARBY_DEVICES group, so it does not add a prompt for users
+      who have already granted NEARBY_WIFI_DEVICES.
+    -->
+    <uses-permission android:name="android.permission.ACCESS_LOCAL_NETWORK" />
     <!-- Signature permission for internal UI shutdown broadcasts -->
     <uses-permission android:name="com.bitchat.android.permission.FORCE_FINISH" />
     <!-- Foreground service and boot permissions for long-running background mesh -->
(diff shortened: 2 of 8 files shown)


moehamade:
Real gap — fixed in c643c90. `ACCESS_LOCAL_NETWORK` is now declared and requested, gated on `SDK_INT >= 37`.

It was missed because the behaviour audit ran against the pre-rebase tree; the Wi-Fi Aware transport arrived with the rebase onto `main`.

One qualification: the review states these connections "will be denied", which I could not confirm. The [docs](https://developer.android.com/privacy-and-security/local-network-permission) frame the feature as LAN access and never address Wi-Fi Aware or link-local IPv6, and these sockets bind to a dedicated Aware `Network`, not the user's subnet. So the permission is declared defensively.

moehamade:
Fixed in ee1ee3ad and 0173b81d.

**Enable-flow gap (real).** The dead end was broader than described — *nobody* could reach the permissions that way, since `shouldRequireWifiAwarePermission()` gates them behind the toggle already being on and it defaults to off. Reproduced on a Pixel 9a (Android 17) and Samsung SM-A366E (Android 16); both needed `adb shell pm grant` before the transport would start. The toggle and Start chip now request first.

**Version gating (real regression from my first fix).** `NEARBY_WIFI_DEVICES` was returned unconditionally despite only existing from API 33, against minSdk 26. On API 26–32 that would request an unknown permission, get denied, and never enable a transport needing no runtime permission there. Both entries are now gated.

**Live grant state.** The callback no longer infers from the result map, which omits already-held permissions.

**A denied `ACCESS_LOCAL_NETWORK` still doesn't block enabling** — the controller starts fine without it (verified on Android 17) and its necessity is unproven, so a denial shouldn't disable a working transport.

Measured on Android 17: grants are independent (`pm grant` of one doesn't cascade) but share the `NEARBY_DEVICES` group, so requesting both yields a single prompt.

On the original finding: with both granted, the Aware data path still fails at `requestNetwork` (`onUnavailable`, both roles, after role reversal) — before `connectAwareClientSocket` runs. No socket attempted, no `SecurityException`/`EPERM`. Discovery works. Not permission-related.

PR #767, opened 2026-07-26
ui: refresh chat branding and text message layout

yet300:
## Summary

- add a compact monochrome pixel-art BitChat icon to the top app bar
- introduce `BitChatBrandButton` while preserving the About and panic-tap actions
- replace the previous `bitchat` title with the brand icon and `/` nickname separator
- redesign regular text messages into a two-row terminal-style layout:
  - sender on the left
  - timestamp and optional PoW difficulty on the right
  - message body on a separate line
- remove the `<@sender>` wrapper from regular text messages
- preserve nickname, URL and geohash interactions
- keep media, system messages and delivery status unchanged
- render active PoW animations inside the same two-row text layout
- remove the unused `singleOrTripleClickable` modifier

## Cross-platform and animation behavior

This intentionally makes Android regular text messages visually diverge from the current iOS-style legacy renderer by removing the `<@sender>` wrapper and moving the timestamp into a separate header row.

Media and system messages retain the legacy format. Active PoW mining uses the same two-row renderer as completed regular text messages; only the body characters animate, so completing mining no longer changes the message layout.

## Design references

- https://dribbble.com/shots/26688133-Bitchat-Brand-Redesign-Concept
- https://dribbble.com/shots/26688187-Bitchat-App-Redesign-Concept

## App Screen
IMAGE 2026-07-26 17:04:37

IMAGE 2026-07-26 17:04:50

## Scope

This PR does not change the application font family.


app/src/main/java/com/bitchat/android/core/ui/component/button/BitChatBrandButton.kt
```diff
@@ -0,0 +1,69 @@
...
+                    delay(MultiClickThreshold)
+                    if (tapCount == 1) {
+                        currentOnClick()
+                    }
+                    tapCount = 0
+                    resetJob = null
+                }
+            }
+        },
+        modifier = modifier.size(32.dp),
```
moehamade: **[change] Touch target drops below the 48dp minimum — and this button carries the panic gesture.**

M3's `IconButton` composes as `modifier.minimumInteractiveComponentSize().size(40.dp)`. Because the caller's modifier is applied *first*, `.size(32.dp)` fixes the incoming constraints at 32dp and `minimumInteractiveComponentSize()` can't expand past it — so the interactive area ends up 32dp, under the 48dp accessibility minimum.

This matters more than usual here: the target it replaces was a `headlineSmall` "bitchat/" text run (much larger), and this button is the entry point for the **triple-tap panic wipe**. A smaller target means more missed taps on a gesture that needs to work under stress.

Keeping the default size and shrinking only the inner `Icon` (already 16dp on line 66) preserves the visual while restoring the target:

```suggestion
        modifier = modifier,
```

app/src/main/java/com/bitchat/android/core/ui/component/button/BitChatBrandButton.kt
```diff
@@ -0,0 +1,69 @@
+package com.bitchat.android.core.ui.component.button
```
moehamade, line 1: **[nit] Package placement.**

`core.ui.component.button` / `core.ui.icon` introduces two new levels of nesting for one file each — while this same PR deletes `core/ui/utils/ModifierExt.kt`, which was the only other file under `core/ui`. Everything else UI-related in this repo lives flat in `com.bitchat.android.ui`.

Not a blocker, but worth picking a direction: either `core/ui` becomes a real shared-component layer (and existing shared bits move into it), or these two files just live in `com.bitchat.android.ui` like their neighbours.

app/src/main/java/com/bitchat/android/core/ui/icon/BitChatIcon.kt
```diff
@@ -0,0 +1,48 @@
...
+                lineTo(0f, 2f)
+                lineTo(1f, 2f)
+                lineTo(1f, 1f)
+                lineTo(2f, 1f)
+                close()
+            }
+        }.build().also { _BitChatIcon = it }
+    }
+
+private var _BitChatIcon: ImageVector? = null
```
moehamade, line 48: **[nit]** This is the standard Figma/ComposeIcons exporter idiom, so no objection if you'd rather keep it consistent with generated output. But since it's hand-written here, `val BitChatIcon: ImageVector by lazy { ... }` expresses the same cache in one line, is thread-safe, and drops the mutable top-level `var`.

app/src/main/java/com/bitchat/android/ui/ChatHeader.kt
```diff
@@ -356,16 +356,18 @@ private fun MainHeader(
             modifier = Modifier.fillMaxHeight(),
             verticalAlignment = Alignment.CenterVertically
         ) {
+            BitChatBrandButton(
+                onClick = onTitleClick,
+                onTripleClick = onTripleTitleClick,
+                contentDescription = stringResource(R.string.app_brand),
```
moehamade: **[change] `app_brand` is `"bitchat/"` — the separator leaks into the accessibility label.**

TalkBack will announce this button as *"bitchat slash, button"*. Two problems:

1. The trailing `/` is now rendered separately as a literal `Text` on line 366, so the one baked into the string resource is redundant *and* audible.
2. A `contentDescription` should describe the **action**, not the branding — a screen-reader user gets no hint that this opens the About sheet.

Worth adding a dedicated string (e.g. `cd_app_brand` = "bitchat, about") rather than reusing `app_brand`. Note this also leaves `app_brand` — currently translated in 10+ locale files — as a string whose only remaining consumer is an accessibility label, which will confuse translators. Either repurpose it explicitly or drop it.

app/src/main/java/com/bitchat/android/ui/ChatUIUtils.kt
```diff
@@ -156,6 +159,112 @@ fun formatMessageAsAnnotatedString(
...
+    currentUserNickname: String,
+    meshService: MeshService,
+    colorScheme: ColorScheme
+): AnnotatedString {
+    val builder = AnnotatedString.Builder()
+    val isDark =
+        colorScheme.background.red + colorScheme.background.green + colorScheme.background.blue < 1.5f
+    val isSelf = message.senderPeerID == meshService.myPeerID ||
+        message.sender == currentUserNickname ||
+        message.sender.startsWith("$currentUserNickname#")
```
moehamade: **[optional] `isDark` + `isSelf` are now computed three times per message.**

This exact pair is duplicated verbatim at lines 249-253 in `formatTextMessageBody`, and `isSelf` a third time in `MessageComponents.kt:421-423`. Three copies of the same three-clause identity check is the kind of thing that drifts.

Two small extensions would collapse all of them:

```kotlin
fun BitchatMessage.isFromSelf(currentUserNickname: String, meshService: MeshService): Boolean =
    senderPeerID == meshService.myPeerID ||
        sender == currentUserNickname ||
        sender.startsWith("$currentUserNickname#")
```

For `isDark`, the `r + g + b < 1.5f` luminance guess is pre-existing, but now that it's copy-pasted three times it's worth noting that `isSystemInDarkTheme()` answers this directly (it does need a `@Composable` context, so it'd be hoisted to the call site and passed in).

app/src/main/java/com/bitchat/android/ui/ChatUIUtils.kt
```diff
@@ -156,6 +159,112 @@ fun formatMessageAsAnnotatedString(
...
+    val builder = AnnotatedString.Builder()
+    builder.pushStyle(
+        SpanStyle(
+            color = Color.Gray.copy(alpha = 0.7f),
+            fontSize = (BASE_FONT_SIZE - 4).sp
+        )
+    )
+    builder.append(timeFormatter.format(message.timestamp))
+    message.powDifficulty?.takeIf { it > 0 }?.let { bits ->
+        builder.append("  ⛨${bits}b")
```
moehamade: **[nit]** Two spaces before the PoW glyph here, but `formatMessageAsAnnotatedString` uses one (`" ⛨${bits}b"`). Since system and media rows still go through the old formatter, the two spacings will be visible side by side in the same list.

```suggestion
        builder.append(" ⛨${bits}b")
```

app/src/main/java/com/bitchat/android/ui/MessageComponents.kt
```diff
@@ -458,9 +399,155 @@ fun MessageItem(
...
+        val metadataText = formatTextMessageMetadata(
+            message = message,
+            timeFormatter = timeFormatter
+        )
+        val bodyText = formatTextMessageBody(
+            message = message,
+            currentUserNickname = currentUserNickname,
+            meshService = meshService,
+            colorScheme = colorScheme
         )
```
moehamade: **[optional] These three formatters re-run on every recomposition of every visible row.**

`appendIOSFormattedContent` compiles two `Regex` objects per call and runs the URL/geohash parsers over the message content, so this isn't free. The pre-existing code had the same issue with a single `formatMessageAsAnnotatedString` call — this PR triples the call sites, which makes it more worth fixing:

```kotlin
val senderText = remember(message.id, currentUserNickname, colorScheme) {
    formatTextMessageSender(message, currentUserNickname, meshService, colorScheme)
}
```

(`message.id` is already used as the `LazyColumn` item key and as the `pointerInput` key below, so it's a consistent identity to key on.)

app/src/main/java/com/bitchat/android/ui/MessageComponents.kt
```diff
@@ -458,9 +399,155 @@ fun MessageItem(
...
+        val haptic = LocalHapticFeedback.current
+        val context = LocalContext.current
+        var senderLayout by remember { mutableStateOf<TextLayoutResult?>(null) }
+        var bodyLayout by remember { mutableStateOf<TextLayoutResult?>(null) }
+
+        Column(
+            modifier = modifier
+                .fillMaxWidth()
+                .padding(vertical = 4.dp),
+            verticalArrangement = Arrangement.spacedBy(2.dp)
```
moehamade: **[question] Vertical spacing is now uneven between message kinds.**

`MessagesList` already applies `verticalArrangement = Arrangement.spacedBy(4.dp)` to the `LazyColumn`. Adding 4dp vertical padding inside the item means:

- text message → text message: **12dp**
- text message → system / image / audio / file message: **8dp**

Was that intentional? If the goal is more breathing room around messages generally, bumping the `spacedBy` in `MessagesList` keeps it uniform across all message types:

```suggestion
        Column(
            modifier = modifier.fillMaxWidth(),
            verticalArrangement = Arrangement.spacedBy(2.dp)
        ) {
```

app/src/main/java/com/bitchat/android/ui/MessageComponents.kt
```diff
@@ -458,9 +399,155 @@ fun MessageItem(
...
+        ) {
+            Row(
+                modifier = Modifier.fillMaxWidth(),
+                horizontalArrangement = Arrangement.SpaceBetween,
+                verticalAlignment = Alignment.CenterVertically
+            ) {
```
moehamade: **[nit]** `SpaceBetween` is redundant here — the sender `Text` below already takes `weight(1f)`, so it consumes all the slack and pushes the metadata to the end regardless. `Arrangement.Start` (the default) would be equivalent and less misleading.

app/src/main/java/com/bitchat/android/ui/MessageComponents.kt
```diff
@@ -458,9 +399,155 @@ fun MessageItem(
...
+                    fontFamily = FontFamily.Monospace,
+                    softWrap = false,
+                    overflow = TextOverflow.Ellipsis,
+                    onTextLayout = { senderLayout = it }
+                )
+                Text(
+                    text = metadataText,
+                    fontFamily = FontFamily.Monospace,
+                    softWrap = false
+                )
```
moehamade: **[change] Two issues with the metadata `Text`.**

**1. Long-press dead zone.** This `Text` has no `pointerInput`, so long-pressing the timestamp no longer opens the message action menu. Previously the entire message was a single long-pressable `Text`. Combined with `SpaceBetween` on the row, the gap between sender and timestamp is dead too. Hoisting the `onLongPress` to the parent `Column` would restore whole-row behaviour in one place.

**2. Possible collision with the delivery-status overlay in DMs.** `MessageItem` overlays `DeliveryStatusIcon` at `Alignment.TopEnd` of the message `Box`, compensating with `endPad = 16.dp` on the content. That padding was sized for the old layout, where the timestamp trailed the *last line* of text. Now it's pinned to the top-right of row 1 — exactly where the overlay sits. `✓✓` at 10sp is roughly 13dp, so it fits inside 16dp but with almost nothing to spare, and `⛨NNb` widens the metadata further.

Both PR screenshots are public chat — could you add one of a DM with a sent message that has a PoW badge? That would settle #2 either way.

app/src/main/java/com/bitchat/android/ui/MessageComponents.kt
```diff
@@ -458,9 +399,155 @@ fun MessageItem(
...
+                            } catch (_: Exception) {
+                            }
+                            haptic.performHapticFeedback(HapticFeedbackType.TextHandleMove)
+                        },
+                        onLongPress = {
+                            haptic.performHapticFeedback(HapticFeedbackType.LongPress)
+                            onMessageLongPress?.invoke(message)
+                        }
+                    )
+                },
```
moehamade: **[change] This is the main refactor I'd ask for: extract the annotated-text tap handling.**

The "resolve tap position → offset → string annotation → dispatch" dance, plus its `var xLayout by remember { mutableStateOf<TextLayoutResult?>(null) }` + `onTextLayout` boilerplate, now exists in **four** places:

| Location | Tags handled |
|---|---|
| `MessageComponents.kt:444` (sender, this PR) | `nickname_click` |
| `MessageComponents.kt:481` (body, this PR) | `nickname_click`, `geohash_click`, `url_click` |
| `MessageComponents.kt:278` (file header) | `nickname_click` |
| `media/ImageMessageItem.kt:70` | `nickname_click` |
| `media/AudioMessageItem.kt:70` | `nickname_click` |

The PR moves *away* from sharing here: it deletes `core/ui/utils/ModifierExt.kt` (a generic, reusable modifier) and inlines ~60 lines of gesture handling in its place. The copies have already diverged — this one handles three tags, the other four handle one — so drift is real rather than hypothetical.

Two extractions, independent of each other:

**1. A component that owns the layout state**, so no call site has to declare `TextLayoutResult` again:

```kotlin
@Composable
fun AnnotatedClickableText(
    text: AnnotatedString,
    modifier: Modifier = Modifier,
    onAnnotationClick: (tag: String, item: String) -> Boolean,  // true = consumed
    onLongPress: (() -> Unit)? = null,
    ...
)
```

Whatever shape you land on needs to preserve the current priority order — `geohash_click` is checked before `url_click`, which matters for content matching both.

**2. Move the behaviour out of the gesture callback.** The geohash-precision branching (lines 496-510) and the URL-scheme normalization (lines 522-533) are plain logic — neither needs composition. As `openGeohash(context, geohash)` and `openUrl(context, rawUrl)` they'd be testable, and `MessageComponents.kt` would stop containing fully-qualified `com.bitchat.android.geohash.GeohashChannelLevel` decisions inside a UI callback.

Happy to take this on in a follow-up if you'd rather keep this PR scoped to the visual redesign — just say so and I'll open one against `main` after this lands.

moehamade (commented):
Nice work — the two-row layout reads well in the screenshots, and a few things in here are genuine improvements over what they replace: swapping `System.currentTimeMillis()` bookkeeping for a cancellable coroutine in the tap counter is more robust, replacing the unused `isDark` parameter of `appendIOSFormattedContent` with `contentColor` is a clean win, and defaulting it to `baseColor` correctly preserves behaviour for the callers that didn't opt in.

I checked out `f28b1fc` and ran `:app:compileDebugKotlin` — **BUILD SUCCESSFUL**, no new warnings attributable to the changed files.

Inline comments cover the specifics. Three things that don't map to a line in this diff:

### 1. Cross-platform parity is the real decision here

This repo's renderer is deliberately iOS-shaped — `formatMessageAsAnnotatedString`'s KDoc says "iOS-style formatting", and the `<@nick>` wrapper, trailing `[HH:mm:ss]` and `#abcd` suffix handling all exist to match bitchat on iOS. Dropping the wrapper and relocating the timestamp makes Android diverge visually from iOS.

That's a maintainer call, not a code-quality one, and it's probably what decides whether this lands. Worth stating explicitly in the PR description rather than leaving it implicit under "redesign", so the discussion happens up front.

### 2. `MatrixEncryptionAnimation.kt` now contradicts itself

`AnimatedMessageDisplay` carries the comment *"Uses IDENTICAL layout structure as normal message for pixel-perfect alignment"*. That's no longer true. While PoW mining runs, the message renders as `<@sender> content [HH:mm:ss]` on one line; the moment `shouldAnimateMessage` flips false, the parent switches to the new two-row layout and the row visibly jumps.

The PR description lists "active PoW animations unchanged" as intentional, and scoping it that way is reasonable — but the file isn't touched, so that stale comment will mislead the next person. Either convert the animated path too, or update the comment.

### 3. Media messages keep the old header style

`ImageMessageItem`, `AudioMessageItem` and the file-message header all still use `formatMessageHeaderAnnotatedString` (`<@nick> ... [time]`). Mixed with the new two-row text rows, the list now speaks two visual languages. Again consistent with the stated scope — flagging it as a follow-up rather than a blocker.

---

**Verified as fine, in case it saves you a round trip:** `message.sender == "system"` matches the codebase convention (`ChatUIUtils.kt:57`, ~25 producers). Dropping `onTap` from the system path loses nothing, since system messages never went through `appendIOSFormattedContent` and so never carried `geohash_click`/`url_click` annotations. And tapping empty space to the right of a short nickname does *not* misfire the handler — `getStringAnnotations` at a collapsed offset equal to the annotation end returns no match.

Nothing here is a correctness bug. The blocking-ish items are the 32dp touch target (panic gesture), the `contentDescription`, and the long-press dead zone; the rest is consistency and cleanup.

app/src/main/java/com/bitchat/android/ui/MessageComponents.kt
yet300: Extracted the annotated-text pointer handling into a reusable AnnotatedClickableText component and migrated the text sender/body plus file, image, and audio headers to it. The component now owns TextLayoutResult and annotation offset resolution while preserving annotation priority and long-press behavior.

Verified with testDebugUnitTest and assembleDebug.

PR #774, opened 2026-07-27
ui: complete the redesign — palette, top bar, composer, About, and a motion pass

callebtc:
## Summary

Continues the redesign started in #767, working from @fhlcreative's proposals ([brand](https://dribbble.com/shots/26688133-Bitchat-Brand-Redesign-Concept), [app](https://dribbble.com/shots/26688187-Bitchat-App-Redesign-Concept)). #767 landed the brand mark and the two-row message layout; this covers the rest — colour, top bar, composer, About, sheets — plus a motion pass across the app.

## Design tokens

New `BitchatPalette` + `LocalBitchatPalette` (`ui/theme/BitchatPalette.kt`).

M3's `ColorScheme` can't express what this app needs: `onSurface` is the terminal green used for chrome, while message bodies have to be neutral. Rather than repurpose M3 slots — which would repaint ~40 unrelated files unpredictably — the palette is explicit, and M3's surface/outline/error slots are derived from it so un-migrated screens still follow along.

`BitchatMotion` holds three duration tokens (120/180/240 ms). Every transition added here uses them with `FastOutSlowInEasing`, and is deliberately asymmetric: ~180 ms in, ~120 ms out. Things arriving deserve attention; things leaving shouldn't be waited on.

Note that `onSurface` is still `#39FF14`. Only files touched here migrated to the palette, so `DebugSettingsSheet`, `LocationNotesSheet`, `SecurityVerificationSheet` and parts of onboarding still render green text on the new surfaces. It reads fine but is inconsistent — a follow-up sweep is warranted.

## Chat messages

- **Neutral body text.** Was `onSurface` (`#39FF14`); now `palette.textPrimary`. Colour is reserved for `@names`, which is what makes a busy channel scannable.
- **Dropped two whole-paragraph bold rules** — "message is mine" and "you were mentioned". Bolding entire paragraphs was the single largest source of visual noise in the old layout.
- **Inline trailing timestamp** instead of a right-aligned column, so a three-word message no longer reserves a full-width row for eight grey characters.
- **Consecutive-message grouping** (`MessageGrouping.kt`): runs from one author omit the repeated `@name`. Peer-ID-authoritative (two people can share a nickname), 5-minute window, never crosses the system / private / channel boundaries.
- **Mention chips** — one contiguous `SpanStyle(background = …)` over `@name#hash`, tinted by *the mentioned peer's* colour so a given name looks identical everywhere it appears.
- **`// Starting Tor…`** replaces `* italic asterisks *` for background narration.
- Gutters 12 → 16.dp. Peer-colour saturation/value raised; **hue derivation is byte-identical to iOS**.

## Top bar

`TopAppBar` replaced with a plain 52.dp `Row`. M3 was injecting a 4.dp pad plus a 12.dp title inset, and the channel badge's `Button` imposed a hidden 58.dp minimum width — the spacing was impossible to specify exactly.

- One icon size (22.dp) and one tap target (44.dp) for every control.
- Globe glyph for geohash channels, Bluetooth for the local mesh. Mesh reads `mesh`, not `#mesh` — it isn't a hashtag channel.
- Connection indicator pinned leftmost in a fixed-width slot, so the cluster never shifts when it appears. **Only shown while connecting**; the steady green "all good" light is gone, since the absence of the indicator already means everything is fine.
- Peer counts roll to their new value, sliding up on join and down on leave.
- Identity cluster is weighted, so a long nickname can no longer push the status icons off screen.
- Bookmark toggle removed — it's one tap away in the Location Channels sheet, and it was occupying a slot in the app's most crowded row.
- Every control uses one press treatment: 0.86 scale on a spring, no background. Ripples do almost nothing on flat terminal chrome.

## Composer

Camera and mic moved *inside* the pill, so the bottom of the screen is one object rather than a pill plus two loose buttons. 52.dp min height, radius derived as `height / 2`.

- Input text is near-white, not terminal green: this is the one surface where you compose rather than read.
- Dedicated neutral-grey `inputOutline` / `inputSurface` tokens. `outline` is green-tinted, and a green cast here made the whole bottom of the screen read as chrome instead of a place to write.
- Typing the first character morphs camera+mic into send via `AnimatedContent` with `SizeTransform`; clearing morphs back.
- Active states are a 20%-alpha tint, not a fill. A solid accent disc beside your own text was the loudest thing on screen.
- One shared `ComposerActionSurface` for all three buttons, so they're pixel-identical while each keeps its own gesture (click / click+long-press / press-and-hold).

## Message arrival motion

New messages slide up 14.dp and fade in on a snappy spring, while the conversation eases out of the way on a softer one — so an arrival lands *onto* a moving list rather than everything travelling at one speed.

`MessageArrivalTracker` is the part that makes this usable. `LazyColumn` composes on demand, so animating on first composition replays the entrance for every old message you scroll back to. It diffs against the set of ids already seen, **during composition rather than in a `LaunchedEffect`** — effects run after the frame composes, by which point a new row has already missed its cue.

Performance: entry runs entirely in `graphicsLayer` (no measure, no layout, no recomposition of content); rows that aren't arriving allocate no `Animatable`, no coroutine and no extra render layer; and bursts over 6 arrivals (history sync, channel switch) skip entry animation, since animating hundreds of rows would spend the whole frame budget on motion nobody asked to see.

## About

Restructured into `INFO` (what it is + how to drive it) and `SETTINGS` tabs, with a sliding indicator, a hero, the six-item capability list, and the security-audit warning card. Also fixes a pre-existing bug where the features card was labelled `APPEARANCE`.

## Sheets

- **Sheets now slide out.** `ModalBottomSheet` only animates itself away when *it* initiates the dismissal; every programmatic close was yanking the composable out of the tree instantly. `LocalSheetDismiss` runs `hide()` to completion first.
- **Channel selection is acknowledged** — the active dot lands on the chosen row, then the sheet leaves.
- **People lists animate** add / remove / reorder. These live inside an outer `LazyColumn` item where nesting a lazy list isn't possible and `animateItem` is unavailable, so `AnimatedRowColumn` uses `LookaheadScope` + `animateBounds` to get the same result for a plain `Column`.
- **Anon capping**: named participants sort ahead of anons regardless of recency and are never trimmed; anons cap at 5 with a fade-out and `"n more…"`. A popular geohash was pushing everyone worth recognising off screen.

## Removed

Proof-of-work header indicator and the matrix mining animation, along with `PoWMiningTracker` which existed only to feed it. The PoW *setting* is untouched.

## Icons

The set is unchanged — an attempt at custom 8-bit glyphs was reverted as too wonky. What did change is sizing and uniformity: one size per context, 44.dp targets throughout. `Icons.Filled.Person` replaces `Groups` for the people affordance (a crowd glyph turns to mush at 22.dp), and six About feature icons use Material approximations of the designer's custom line art — **worth requesting SVG exports from @fhlcreative** to finish that off.

## Strings

New keys for the About tabs, how-to rows, warning card, sheet sections and Tor routing. Where a role and length matched, existing already-translated keys were reused rather than minting English-only ones (`about_offline_mesh_*`, `bookmarked`, `channels`, `disable_location_services`, `about_emergency_title`), and 162 genuinely dead entries were pruned across 30 locale files. Remaining `MissingTranslation` warnings are new UI copy, handled the same way #767 did.

## Testing

`./gradlew test lintDebug assembleDebug` all green. **279 unit tests, 0 failures** — 46 new:

- `MessageGroupingTest` (16) — identity, the 5-minute window, out-of-order timestamps, system/private/channel boundaries, spacing.
- `MessageArrivalTrackerTest` (11) — first load, empty first load, re-presenting an identical list, recycling an already-seen row, partial batches, the burst-cap boundary, `/clear` pruning, a channel switch where size is unchanged but every id differs.
- `GeohashAnonOrderingTest` (8) — anon classification (`anon`/`anon1234` yes, `anonymous`/`anonracer` no), two-tier ordering, cap arithmetic.
- `ChatUIUtilsTest` (+11) — inline timestamps, neutral body colour, the no-bold regression guard, mention-chip contiguity and tinting, `//` system format, iOS hue parity.

Lint was diffed against `main` in a worktree: no new `UnusedResources`, no new rule categories, one `TypographyEllipsis` fixed.

## Needs a device

The motion can't be verified off-hardware. Worth a look at: a rapid burst in a busy geohash (confirming the arrival cap engages), light theme, teleported state, the DM sheet, a 15-char nickname, and an RTL locale for the header `Row`.

## Cross-platform

Extends the divergence #767 accepted: inline timestamps, `//` system messages, grouping and neutral body text all differ from iOS now. Peer-colour **hue** stays byte-identical, so a given peer is the same hue on both platforms.


fhlcreative:
So excited to see you working on this @callebtc ! 🙌

I thought it might be useful to provide the actual Figma file for any assets you might need: 
https://www.figma.com/design/Ffo3e0mBbgIPnDMvE225gX/Bitchat-UI?node-id=0-1&t=FfUy12hL9bkalCiJ-1

yet300:
@callebtc After teleporting, you can no longer bookmark your current location. It used to be in the app bar, but I think you could add the currently open Location Channels to the list (right now, only nearby locations are displayed as selected).

PR #823, opened 2026-07-29
feat: animated 3D globe geohash picker with borders and cities

callebtc:
## Summary

Replaces the Leaflet/WebView geohash picker with a fully offline, native Compose **3D globe** that matches the app's dark terminal aesthetic:

- **Animated orthographic globe** rendered on a Compose Canvas: starfield, atmosphere glow, shaded ocean sphere, graticule, and vector continents from bundled public-domain Natural Earth 110m data
- **Geohash cells projected onto the sphere** — selected cell glows in the app green with Geist Mono labels, level + coverage readout (`City • ~4.9 km`), pulsing crosshair
- **Gestures**: drag to spin (inertial fling), pinch to zoom (precision follows zoom automatically), tap to focus, double-tap to zoom, cinematic 1.4s fly-in intro, haptic tick on cell change
- **Country borders + cities** (Natural Earth admin-0 boundaries & 50m populated places, bundled as 155KB of GeoJSON): capitals as accent dots, zoom-tiered city labels that fade in at regional zoom
- **Fully off-grid**: removes the WebView, Leaflet CDN, and raster tile dependencies entirely

Same Activity contract as before (`initial_geohash` → `result_geohash`); the only call site in `LocationChannelsSheet` is untouched.

## Implementation notes

- Horizon-clipped polygon fills: projected rings are split into front-facing runs, closed along limb arcs; strokes never draw through the far side
- All path geometry is clipped to an expanded viewport rect before rasterization — Skia loses precision for path coordinates beyond ~32767px, which happens at high globe zoom
- Land/border/city GeoJSON is parsed once on IO and cached; rendering allocates no per-frame bitmaps

## Verification

Screenshot-verified on an Android 15 emulator (1080×1920) across whole-globe, region, province, city, mid-ocean, antimeridian, dark and light themes, plus drag/tap/double-tap/buttons and Select result flow. The picker previously required network for tiles; it now renders identically with airplane mode on.


callebtc:
Addressed the polar-geohash review comment in 03de6f35: the latitude clamp is now ±89° (from ±80°), so opening or tapping polar geohashes centers on the requested cell and syncSelection() encodes the right geohash. Screenshot-verified with an 87°N p2 cell and an 85°S p6 cell.

PR #827, opened 2026-07-29
Fix stale peer lifecycle cleanup

a1denvalu3:
## Summary

- keep lifecycle cleanup callbacks active when stale peer IDs are replaced
- suppress only the intermediate peer-list update during replacement
- remove inbound as well as outbound references when deleting a peer from the mesh graph
- add regression coverage for both cleanup paths

## Why

PeerManager used one flag for both UI list updates and the onPeerRemoved lifecycle callback. Stale same-nickname replacement disabled that flag to avoid an intermediate list update, but also skipped cleanup of authenticated state, gossip, mesh graph routes, and Noise sessions. This could retain a stale route across process restart and cause post-restart direct messages to time out.

## Validation

- ./gradlew :app:testDebugUnitTest --tests com.bitchat.PeerManagerTest --tests com.bitchat.android.services.meshgraph.MeshGraphServiceTest
- two-device Mesh Lab session_recovery: pass, including bidirectional DMs before and after process restart
- two-device Mesh Lab full suite: pass, including DM, broadcast, file/media transfer, session recovery, and identity reset

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/PeerManager.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/PeerManager.kt
@@ -335,7 +335,7 @@ class PeerManager {
         
         // Remove stale peer IDs
         stalePeerIDs.forEach { stalePeerID ->
-            removePeer(stalePeerID, notifyDelegate = false)
+            removePeer(stalePeerID, notifyPeerList = false)
         }
         
         // Check if this is a new peer announcement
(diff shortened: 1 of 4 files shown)


a1denvalu3:
## Additional fix explanation

Each peer publishes a set of neighbors it claims to see. Conceptually, the graph might contain:

  A → B
  B → A, C
  C → B

  When removing peer B:

  - “Outbound references” are B’s own announcement: B → A, C.
  - “Inbound references” are other peers mentioning B: A → B and C → B.

  Previously, removePeer("B") deleted only B’s own announcement. A and C could still list B as a neighbor, so B
  remained as a ghost node with stale edges in the graph.

  The fix removes B from every remaining peer’s neighbor set too:

  Before: A → B, B → A/C, C → B
  After:  A → nothing, C → nothing

  This prevents routing logic or graph snapshots from continuing to consider a removed peer reachable.

a1denvalu3:
I also ran the full mesh-lab suite with two phones, multiple times. Tests all pass now and spurious failures are gone.

PR #828, opened 2026-07-29
fix(hotspot): never disturb another app's Wi-Fi Direct group without consent

moehamade:
Replaces #811.

#811 set out to fix two bugs and grew to 29 commits and ~3,100 added lines under automated review pressure — an explicit state machine, ownership-evidence tracking, observation counting, and channel-recovery verification. Most of that machinery existed to answer one question without asking the user: *can we prove the current P2P group is ours before removing it?* This PR asks the user instead, which makes the machinery unnecessary. Same protection, proportionate cost: **8 files, ~430 lines added** against the versions that shipped in #808.

## The two bugs (unchanged from #811's original scope)

**1. Stopping the hotspot tore down other apps' Wi-Fi Direct sessions.** `stopHotspot()` called `removeGroup()` unconditionally, and removal is *device-scoped* — it removes whatever group exists, including one owned by Cast, Android Auto, or Quick Share. The preflight built to protect a foreign group failed through this exact path, disconnecting the session it had just refused to touch.

Fix: a single `createdGroup` flag, set once our own `createGroup` command is accepted. Stop only issues `removeGroup()` when it's set; otherwise the channel is closed and the group left alone.

**2. A Wi-Fi Aware restart request could be swallowed.** `restartIfStillEnabled()` coalesces on an in-flight flag; a request arriving while an earlier loop was burning attempts against the hotspot hold lost the CAS and was dropped, leaving the mesh down despite the user's setting. Fix: requests are recorded before coalescing and re-checked after each pass.

The Aware hold is also now a **counted, once-releasable lease** (`acquireHotspotLease()`), so overlapping share sessions cannot release each other's claim, and duplicate teardown paths cannot double-release.

## Consent instead of ownership proof

If a Wi-Fi Direct group we did not record creating is active at start, the app no longer guesses — it shows a dialog: *"Starting the hotspot will disconnect the current Wi-Fi Direct connection. This may interrupt Cast, Android Auto, Quick Share, or another nearby connection."* Confirming retries the start with removal authorized; cancelling leaves everything untouched.

Consent is **bound to the group it was given for**: the conflicting group's name travels through the dialog, and the policy only authorizes removing a group with exactly that name. A group that appeared after confirmation — or swapped in mid-BUSY-retry — re-prompts instead of riding on stale approval.

Ownership is **exact-name only**: the `DIRECT-BC-` prefix heuristic is gone. A prefix match is not proof — this device can be connected to *another phone's* bitchat group — and a group we can't prove ours simply goes through the dialog once. The recorded name is not updated from a group we don't host (client role), and is not overwritten while an old group of ours may still exist — either would let a later check misclassify a group.

## Also fixed

- `SecurityException` guards on the `removeGroup()` call sites that run after permission checks (permission revoked mid-session crashed instead of failing cleanly).
- `stopHotspot(onTeardownComplete)`: the ViewModel releases the Aware lease only after the framework acknowledges our group's removal, instead of immediately — restoring Aware while our group is still up recreates the NAN/P2P contention the lease exists to prevent. An idempotent 10s fallback close means a dropped acknowledgement cannot pin the mesh down for the life of the process.
- Confirm/cancel handlers guard on the `ConfirmDisconnect` state, so a tap landing through a screen transition cannot tear down a just-confirmed session.

## Ownership is re-checked, not remembered

`createdGroup` is reconciled against every group snapshot rather than set once, and no snapshot is trusted until the group is confirmed to be the one we created. `isGroupOwner` cannot make that call alone — it reports that *this device* hosts the group, which is equally true of an autonomous group another app created here — so identity comes from the network name: above Q the one we chose, below Q the first name the framework reported for our own successful creation, fixed from then on. Without that, a replacement group's name was persisted as `ownedGroupName` and the next start removed it silently, since ownership is now an exact-name match. If our group disappears and another app claims P2P while the sharing screen is open, stop no longer believes the replacement is ours. Reconciliation waits until our group has been named, since a null snapshot during formation says nothing — and losing the flag to a transient null is the safe direction, because the group we created is then merely left behind for the next start to recognize and remove.

## Deliberately not handled

Under consent semantics these are accepted rather than engineered around; listing them so the choice is visible:

- **Stop during the ~100 ms create-acceptance window**: the group may land orphaned. The next start recognizes it by its recorded name and cleans it up silently (device-verified below); at worst the dialog appears once.
- **Orphans from builds predating the ownership record**: one confirmation dialog, then recorded and silent thereafter.
- **Hotspot outliving the 30-minute wake-lock timeout**: pre-existing behavior, unchanged here.

Each has a rare trigger, a bounded consequence, and a self-healing follow-up — that trade is the point of this PR.

## Screenshots

Captured on a Pixel 9a (Android 16, API 37). Hosted on a [standalone branch](https://github.com/moehamade/bitchat-android/tree/assets/hotspot-consent-screenshots) rather than committed here, so this change adds no binaries to the project history.

| Consent dialog | Hosting | After consent |
|---|---|---|
| Dialog: Disconnect current Wi-Fi Direct connection? | Hotspot Active with QR code, SSID and password | Hotspot Active after confirming replacement |

A group the app cannot prove is its own produces the dialog on the left — nothing is touched until the user answers. Cancel returns to the intro screen with the group intact.

## Verification

- Full unit suite green (519 tests), including new policy tests (consent action, name-bound consent, exact-name ownership) and the lease counting rules.
- Cross-checked against all 34 Codex review threads from #811: the finding classes that apply to this design (dropped teardown acknowledgement pinning the radio; consent not bound to a group identity; ownership marker overwritten while the old group survives; ownership not re-checked before removal) are addressed above; the rest concern state-machine code this PR does not contain.

**Device-verified on a Pixel 9a (Android 16, API 37):**

| Scenario | Result |
|---|---|
| Start with no group present | Hosting in ~0.5 s; group created, web server on :9999 |
| Stop | `Group removed successfully` → `P2P channel closed` → `restoring Wi-Fi Aware` → mesh service re-instantiated. `mGroup null`, no orphan, correct ordering every cycle |
| Aware hold while hosting | `Not starting Wi-Fi Aware: held down for the hotspot` — the lease blocks Aware restarts for the whole session |
| Own orphan after process death (killed while hosting) | `Removing stale group 'DIRECT-BC-KB9XDUKA' before creating` → removed → new group. No dialog: recognized by its recorded name |
| Unrecorded group present (marker erased so the live group is indistinguishable from another app's) | Consent dialog shown; logcat confirms **no `removeGroup()` issued** — only our own channel closed |
| Cancel | Returns to intro, nothing removed, Aware lease released |
| Disconnect and start | Start proceeds; new group created (one `BUSY` absorbed by the 1 s backoff) |
| Transient null snapshot during formation | Ignored by ownership reconciliation — `createdGroup` retained, start completes |

Limit of that simulation, stated plainly: a genuinely foreign group cannot be created from a test harness, so the "unrecorded group" case used one of bitchat's own orphans with the ownership marker erased. That proves the decision path and that `removeGroup()` is not issued, but the group still vanished during the dialog — `Channel.close()` is uid-scoped, so closing our channel reaps our own app's leftover group. A group owned by another package is outside that scope and is therefore untouched by either code path. Verifying that end-to-end needs a real Cast/Android Auto/Quick Share session.

- Known gaps, tracked as follow-up rather than blocking: no unit tests for the ViewModel consent flow or the `WifiAwareController` restart-coalescing paths (both need refactoring for testability); the three new dialog strings are English-only while `cancel` is translated in ~20 locales, so a non-English user sees a mixed-language dialog (observed with the app locale set to Italian).

### DIFF
--- a/app/src/main/java/com/bitchat/android/hotspot/HotspotManager.kt
+++ b/app/src/main/java/com/bitchat/android/hotspot/HotspotManager.kt
@@ -63,11 +63,16 @@ class HotspotManager(private val context: Context) {
     private var hasNotifiedStarted = false // Track if we've notified the callback
     private var isReceiverRegistered = false // Track receiver registration to prevent leaks
 
-    // Set once our own createGroup command is accepted. stopHotspot() only calls
-    // removeGroup() when this is true: removal is device-scoped, so issuing it with
-    // nothing of ours on the framework could only tear down another app's session
-    // (Cast, Android Auto, Quick Share).
+    // Set once our own createGroup command is accepted, and re-checked against every
+    // group snapshot afterwards. stopHotspot() only calls removeGroup() when this is
+    // true: removal is device-scoped, so issuing it when the group on the framework
+    // is not ours could only tear down another app's session (Cast, Android Auto,
+    // Quick Share).
     private var createdGroup = false
+
+    // Framework-reported name of the group this session hosts, once known. Null while
+    // the group is still forming, when a null snapshot carries no information.
+    private var hostedGroupName: String? = null
 
     // Name of the foreign group the user explicitly agreed to disconnect, or null.
     // Consent is per-group: a group with a different name asks again.
@@ -199,6 +204,7 @@ class HotspotManager(private val context: Context) {
 
         val hadOwnGroup = createdGroup
         createdGroup = false
+        hostedGroupName = null
 
         if (staleChannel != null && hadOwnGroup) {
             try {
@@ -557,40 +563,105 @@ class HotspotManager(private val context: Context) {
 
         try {
             wifiP2pManager?.requestGroupInfo(ch) { group ->
-                if (group != null) {
-                    currentGroup = group
+                // A reply arriving after the hotspot stopped must not revive any
+                // state the stop just cleared.
+                if (channel !== ch) return@requestGroupInfo
 
-                    // Update saved credentials if using system-generated ones
-                    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) {
-                        savedSsid = group.networkName
-                        savedPassword = group.passphrase
-                    }
+                reconcileGroupOwnership(group)
 
-                    // Authoritative name straight from the framework. Only recorded
-                    // for a group we host: in the client role the group is another
-                    // app's, and recording its name would make the next run treat a
-                    // foreign orphan as ours.
-                    if (group.isGroupOwner) {
-                        group.networkName?.let { ownedGroupName = it }
-                    }
-
-                    // Notify callback on FIRST successful group info retrieval
-                    if (!hasNotifiedStarted) {
-                        hasNotifiedStarted = true
-                        Log.d(TAG, "Group info received, notifying callback")
-                        callback?.onHotspotStarted()
-                    } else {
-                        // Subsequent updates
-                        callback?.onConnectionInfoUpdated(getConnectionInfo())
-                    }
-                } else {
+                if (group == null) {
                     Log.w(TAG, "requestGroupInfo returned null group")
+                    return@requestGroupInfo
+                }
+
+                if (!isOurHostedGroup(group)) {
+                    // Someone else's group is on the radio. Reading anything from it
+                    // — its name, its credentials, its client count — would report
+                    // another app's session as our hotspot, and recording its name
+                    // would let the next start remove it without asking.
+                    Log.w(
+                        TAG,
+                        "Observed group '${group.networkName}' is not the one we created"
+                    )
+                    return@requestGroupInfo
+                }
+
+                currentGroup = group
+
+                // Update saved credentials if using system-generated ones
+                if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) {
+                    savedSsid = group.networkName
+                    savedPassword = group.passphrase
+                }
+
+                // Authoritative name straight from the framework, for the group we
+                // just confirmed is ours.
+                group.networkName?.let {
+                    hostedGroupName = it
+                    ownedGroupName = it
+                }
+
+                // Notify callback on FIRST successful group info retrieval
+                if (!hasNotifiedStarted) {
+                    hasNotifiedStarted = true
+                    Log.d(TAG, "Group info received, notifying callback")
+                    callback?.onHotspotStarted()
+                } else {
+                    // Subsequent updates
+                    callback?.onConnectionInfoUpdated(getConnectionInfo())
                 }
             }
         } catch (e: SecurityException) {
             Log.e(TAG, "Wi-Fi permission was revoked while reading group info", e)
             failStartup("A required Wi-Fi or local network permission was revoked. Grant it and try again.")
         }
+    }
+
+    /**
+     * Is this snapshot the group this session created?
+     *
+     * `isGroupOwner` cannot answer that on its own: it reports that *this device*
+     * hosts the group, which is equally true of an autonomous group another app
+     * created here. Above Q we chose the network name, so it identifies our group
+     * exactly. Below Q the framework names it, and the first snapshot after our own
+     * createGroup succeeded is the only evidence available — after that the name is
+     * fixed, and a group answering to a different one is not ours.
+     */
+    private fun isOurHostedGroup(group: WifiP2pGroup): Boolean {
+        if (!group.isGroupOwner) return false
+        val name = group.networkName ?: return false
+
+        hostedGroupName?.let { return name == it }
+
+        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            name == savedSsid
+        } else {
+            true
+        }
+    }
+
+    /**
+     * Keep [createdGroup] honest about what is actually on the framework.
+     *
+     * Our group can disappear without us — Wi-Fi toggled, another app issuing its own
+     * device-scoped removeGroup(), a driver reset — and another app can then create
+     * one in its place. Believing the group present is still ours would make stop
+     * remove that replacement, the exact disruption consent exists to prevent.
+     *
+     * Reconciled only once the framework has named our group: before that a null
+     * snapshot means the group is still forming, not that it is gone. Losing the flag
+     * to a transient null is safe in a way that keeping it is not — the group we
+     * created is then left behind, and the next start recognises it by name and
+     * removes it silently.
+     */
+    private fun reconcileGroupOwnership(group: WifiP2pGroup?) {
+        val hosted = hostedGroupName ?: return
+
+        val stillOurs = group != null && isOurHostedGroup(group)
+        if (createdGroup && !stillOurs) {
+            Log.w(TAG, "Group '$hosted' is no longer ours; leaving what is present alone")
+        }
+        createdGroup = stillOurs
     }
 
     /**
--- a/app/src/main/java/com/bitchat/android/hotspot/HotspotManager.kt
+++ b/app/src/main/java/com/bitchat/android/hotspot/HotspotManager.kt
@@ -78,6 +78,12 @@ class HotspotManager(private val context: Context) {
     // Consent is per-group: a group with a different name asks again.
     private var confirmedReplacementName: String? = null
 
+    // stopHotspot() can be reached again while its group query/removal is still in
+    // flight. Later callers wait for that same teardown instead of releasing the
+    // Wi-Fi Aware lease early.
+    private var teardownInProgress = false
+    private val teardownCallbacks = mutableListOf<() -> Unit>()
+
     // Saved credentials for reconnection
     private var savedSsid: String? = null
     private var savedPassword: String? = null
@@ -191,6 +197,12 @@ class HotspotManager(private val context: Context) {
     fun stopHotspot(onTeardownComplete: (() -> Unit)? = null) {
         Log.d(TAG, "Stopping hotspot")
 
+        onTeardownComplete?.let(teardownCallbacks::add)
+        if (teardownInProgress) {
+            Log.d(TAG, "Teardown already in progress; chaining completion")
+            return
+        }
+
         isStarting = false
         hasNotifiedStarted = false
 
@@ -203,40 +215,27 @@ class HotspotManager(private val context: Context) {
         channel = null
 
         val hadOwnGroup = createdGroup
+        val expectedGroupName = hostedGroupName ?: if (
+            Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q
+        ) {
+            savedSsid
+        } else {
+            null
+        }
         createdGroup = false
         hostedGroupName = null
 
+        var teardownAction: (() -> Unit)? = null
         if (staleChannel != null && hadOwnGroup) {
-            try {
-                wifiP2pManager?.removeGroup(staleChannel, object : ActionListener {
-                    override fun onSuccess() {
-                        Log.d(TAG, "Group removed successfully")
-                        // Nothing of ours is left for a later run to clean up.
-                        ownedGroupName = null
-                        closeChannel(staleChannel)
-                        onTeardownComplete?.invoke()
-                    }
-                    override fun onFailure(reason: Int) {
-                        Log.w(TAG, "Failed to remove group: $reason")
-                        closeChannel(staleChannel)
-                        onTeardownComplete?.invoke()
-                    }
-                })
-            } catch (e: SecurityException) {
-                // Revoked mid-session. Closing the channel still detaches this app's
-                // binder, which asks the framework to drop our group with it.
-                Log.e(TAG, "Wi-Fi permission was revoked while removing the group", e)
-                closeChannel(staleChannel)
-                onTeardownComplete?.invoke()
+            teardownInProgress = true
+            teardownAction = {
+                removeOwnGroupIfStillPresent(staleChannel, expectedGroupName)
             }
         } else if (staleChannel != null) {
             // This session created nothing, so there is nothing of ours to remove.
             // removeGroup() here is exactly the bug this change fixes: device-scoped
             // removal would disconnect whatever group another app has running.
             closeChannel(staleChannel)
-            onTeardownComplete?.invoke()
-        } else {
-            onTeardownComplete?.invoke()
         }
 
         // Release locks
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/bitchat/android/hotspot/HotspotManager.kt
```diff
@@ -177,19 +197,40 @@ class HotspotManager(private val context: Context) {
...
-                }
-                override fun onFailure(reason: Int) {
-                    Log.w(TAG, "Failed to remove group: $reason")
-                    closeChannel(staleChannel)
-                }
-            })
+        val hadOwnGroup = createdGroup
+        createdGroup = false
+
+        if (staleChannel != null && hadOwnGroup) {
```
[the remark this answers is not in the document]
moehamade: Confirmed and fixed in 5859ac2. `createdGroup` is now reconciled against every group snapshot rather than set once: still ours only while the observed group is the one this session created. Reconciliation deliberately waits until our group has been named by the framework, because a null snapshot during formation says nothing about a group that has not appeared yet — device-verified that a transient null mid-formation no longer drops ownership. Losing the flag to a transient null is the safe direction: the group we created is then left behind for the next start to recognise by name and remove silently, rather than a stranger being removed.

app/src/main/java/com/bitchat/android/hotspot/HotspotManager.kt
```diff
@@ -481,8 +578,16 @@ class HotspotManager(private val context: Context) {
...
-                    // Authoritative name straight from the framework
-                    group.networkName?.let { ownedGroupName = it }
+                    // Authoritative name straight from the framework. Only recorded
+                    // for a group we host: in the client role the group is another
+                    // app's, and recording its name would make the next run treat a
+                    // foreign orphan as ours.
+                    if (group.isGroupOwner) {
+                        group.networkName?.let {
+                            hostedGroupName = it
+                            ownedGroupName = it
```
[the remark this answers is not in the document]
moehamade: Valid — fixed in 5859ac2. You are right that `isGroupOwner` reports the local device's P2P role, so it is equally true of an autonomous group another app created here, and clearing the immediate flag did not stop the replacement being authorised through the persisted marker. That mattered more in this PR than before it, because ownership is now an exact-name match rather than a prefix heuristic.

Snapshots are no longer read at all until the group is confirmed to be the one we created (`isOurHostedGroup`): above Q by the network name we set in the `WifiP2pConfig`, below Q by the first name the framework reported for our own successful `createGroup`, fixed from then on. A group answering to any other name is not adopted as `ownedGroupName`, its credentials are not copied into the shown SSID/passphrase below Q, and its client count is not reported as ours.

app/src/main/java/com/bitchat/android/hotspot/HotspotStartupPolicy.kt
```diff
@@ -42,29 +44,38 @@ internal object HotspotStartupPolicy {
...
         existingGroupName: String?,
-        ownedGroupName: String?
+        ownedGroupName: String?,
+        confirmedGroupName: String?
     ): StartAction = when {
         p2pState == WifiP2pManager.WIFI_P2P_STATE_DISABLED -> StartAction.Fail(P2P_DISABLED_MESSAGE)
         existingGroupName == null -> StartAction.Create
         isOurs(existingGroupName, ownedGroupName) -> StartAction.RemoveStaleGroupThenCreate
-        else -> StartAction.Fail(FOREIGN_GROUP_MESSAGE)
+        existingGroupName == confirmedGroupName -> StartAction.RemoveStaleGroupThenCreate
```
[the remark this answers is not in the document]
moehamade: Checked this one carefully and I am declining it — the mechanism is real but the consequence is not reachable in a way that violates consent.

**The plausible case is within what the user authorized.** The dialog names a *kind* of connection, not an opaque identity: "will disconnect the current Wi-Fi Direct connection. This may interrupt Cast, Android Auto, Quick Share, or another nearby connection." If a Cast session ends and Cast reconnects under the same app-chosen SSID, removing it is exactly what the user asked for when they tapped "Disconnect and start". Re-prompting for the same logical connection would be a worse experience, not a safer one.

**The violating case needs two different apps to collide on the same app-chosen string.** On Android 10+ the name is selected by the creating app, so for stale consent to authorize the wrong session, app A's group must end and app B must independently pick byte-identical `networkName`. There is no mechanism that makes that likely.

**The suggested remedy does not discriminate in the case that matters.** For a foreign *autonomous* group, this device is the group owner, so `WifiP2pGroup.getOwner()` returns our own device and is identical across the old and replacement group — a MAC-based fingerprint adds nothing precisely where it would be needed. `getNetworkId()` is not specified as a per-session handle, and a passphrase is as reusable as the name for a reconnecting app. I could not find a stronger identity that `requestGroupInfo()` actually exposes for this case.

Scope is also already bounded: consent lives for one start attempt (cleared on confirm, cancel, stop, error, and any failure), and the retry window is ~15 s.

Happy to revisit with either of: a concrete mechanism by which two distinct apps produce the same `networkName`, or a documented `WifiP2pGroup` field that is stable within a session and distinct across sessions. Without one of those, adding a fingerprint here is mechanism against an unreachable case — which is the failure mode this PR exists to correct (see #811, 29 commits of exactly that).

app/src/main/java/com/bitchat/android/hotspot/HotspotManager.kt
```diff
@@ -177,19 +202,41 @@ class HotspotManager(private val context: Context) {
...
-                override fun onFailure(reason: Int) {
-                    Log.w(TAG, "Failed to remove group: $reason")
-                    closeChannel(staleChannel)
-                }
-            })
+        val hadOwnGroup = createdGroup
+        createdGroup = false
+        hostedGroupName = null
+
+        if (staleChannel != null && hadOwnGroup) {
```
callebtc, line 229: [P1] Revalidate ownership immediately before removal

`createdGroup` reflects only the last poll. If another app replaces this group before `stopHotspot()` runs, `hadOwnGroup` remains true and this device-scoped `removeGroup()` disconnects the replacement without consent. The reconciliation added for the earlier review does not synchronize this final removal. Re-request group info on `staleChannel` and compare it with the captured hosted name immediately before removal.

app/src/main/java/com/bitchat/android/hotspot/HotspotManager.kt
```diff
@@ -503,6 +617,53 @@ class HotspotManager(private val context: Context) {
...
+     * remove that replacement, the exact disruption consent exists to prevent.
+     *
+     * Reconciled only once the framework has named our group: before that a null
+     * snapshot means the group is still forming, not that it is gone. Losing the flag
+     * to a transient null is safe in a way that keeping it is not — the group we
+     * created is then left behind, and the next start recognises it by name and
+     * removes it silently.
+     */
+    private fun reconcileGroupOwnership(group: WifiP2pGroup?) {
+        val hosted = hostedGroupName ?: return
```
callebtc: [P2] Reject mismatched snapshots during formation

While `hostedGroupName` is null, reconciliation ignores even a non-null group whose name differs from `savedSsid` on Android Q+. That leaves `createdGroup` true; polling eventually times out and feeds the foreign group into the unsafe stop path. Ignore null snapshots during formation, but demote ownership immediately when a known name does not match the configured SSID.

app/src/main/java/com/bitchat/android/hotspot/HotspotViewModel.kt
```diff
@@ -134,10 +189,34 @@ class HotspotViewModel(application: Application) : AndroidViewModel(application)
...
+        if (manager == null) {
+            lease.close()
+            return
+        }
+
+        // Close the lease only when the manager has finished removing our group.
+        // Restoring Wi-Fi Aware earlier recreates the NAN/P2P radio contention the
+        // lease exists to prevent. The lease is idempotent, so a duplicate or late
+        // completion cannot release a newer session's claim.
+        manager.stopHotspot { lease.close() }
```
callebtc, line 209: [P1] Chain repeated teardown completion

`HotspotManager.failStartup()` first calls `stopHotspot()` without a completion callback, starting asynchronous removal, and then `onError` reaches this second stop. Because the first call already cleared the manager's `channel`, this call reports completion immediately and releases the Wi-Fi Aware lease while removal is still running. Chain this callback onto the in-progress teardown, or report the error only after the first stop completes.

app/src/main/java/com/bitchat/android/hotspot/HotspotManager.kt
```diff
@@ -177,19 +202,41 @@ class HotspotManager(private val context: Context) {
...
+        createdGroup = false
+        hostedGroupName = null
+
+        if (staleChannel != null && hadOwnGroup) {
+            try {
+                wifiP2pManager?.removeGroup(staleChannel, object : ActionListener {
+                    override fun onSuccess() {
+                        Log.d(TAG, "Group removed successfully")
+                        // Nothing of ours is left for a later run to clean up.
+                        ownedGroupName = null
```
callebtc: [P2] Do not clear a newer ownership marker

This asynchronous callback unconditionally clears the shared preference. The UI permits another session to start before it arrives, so a newer `HotspotManager` may already have recorded its group; the old callback then erases that marker. Capture the stopped group name and clear only when the stored value still matches it.

callebtc (changes requested):
The current implementation still has two ownership/consent holes and two teardown races. CI is green, but these asynchronous paths are not covered by the added tests.

app/src/main/java/com/bitchat/android/hotspot/HotspotManager.kt
moehamade: Fixed in b6ad897. `removeOwnGroupIfStillPresent()` re-queries `requestGroupInfo` on the stale channel and gates `removeGroup()` on `HotspotStartupPolicy.isExpectedHostedGroup(...)`, so the removal acts on a fresh snapshot rather than a poll up to a second old. A snapshot that is not ours closes the channel and completes teardown without touching the group.

Verified on a Pixel 9a (Android 16, API 37) by disabling Wi-Fi while hosting:
```
HotspotMgr: Current group 'null' is not ours ('DIRECT-BC-KBEBBQCF'); leaving it alone
HotspotMgr: P2P channel closed
```
No device-scoped `removeGroup()` was issued on that path.

app/src/main/java/com/bitchat/android/hotspot/HotspotManager.kt
moehamade: Fixed in b6ad897. `reconcileGroupOwnership()` now falls back to the configured `savedSsid` when `hostedGroupName` is still unset, and only excuses a *null* snapshot (`if (hostedGroupName == null && group == null) return`). A non-null group naming something else is treated as positive evidence and clears `createdGroup`, so the 15 s formation timeout can no longer hand a foreign group to the stop path.

Below Q the expected name is unknown until the framework reports it, so that window stays as documented — the framework names the group there and no earlier identity exists to compare against.

app/src/main/java/com/bitchat/android/hotspot/HotspotViewModel.kt
moehamade: Fixed in b6ad897. `stopHotspot()` queues completions into `teardownCallbacks` and returns early when `teardownInProgress` is set, so the second stop — `failStartup()` stops the manager, then reports the error, which stops it again through the ViewModel — chains onto the running teardown instead of reporting completion against an already-detached channel.

Verified on a Pixel 9a by disabling Wi-Fi while hosting:
```
HotspotViewModel: Hotspot failed: Wi-Fi Direct is unavailable...
HotspotMgr: Teardown already in progress; chaining completion
HotspotMgr: Current group 'null' is not ours; leaving it alone
HotspotMgr: P2P channel closed
WifiAwareController: Hotspot finished; restoring Wi-Fi Aware if enabled
```
The Aware lease is released only after teardown resolves, not while removal is in flight.

app/src/main/java/com/bitchat/android/hotspot/HotspotManager.kt
moehamade: Fixed in b6ad897. The removal callback now goes through `clearOwnedGroupNameIfMatches(expectedGroupName)`, and `HotspotStartupPolicy.shouldClearOwnedGroupName()` clears only when the stored marker still equals the group this teardown actually removed. A late callback from a previous session can no longer erase the marker a newer session recorded, so that session's own orphan stays recognisable and is removed silently rather than prompting the user to disconnect our own dead group.

moehamade:
All four review findings are addressed by b6ad897 and verified independently.

| # | Finding | Closure |
|---|---|---|
| P1 | Revalidate ownership before removal | `removeOwnGroupIfStillPresent()` gates `removeGroup()` on a fresh snapshot via `isExpectedHostedGroup()` |
| P1 | Chain repeated teardown completion | `teardownInProgress` + `teardownCallbacks`; the second stop chains instead of completing early |
| P2 | Reject mismatched snapshots during formation | `reconcileGroupOwnership()` falls back to `savedSsid` and excuses only null snapshots |
| P2 | Do not clear a newer ownership marker | `shouldClearOwnedGroupName()` clears only the group this teardown removed |

**Verification**

- 521 unit tests pass locally, including the two new `HotspotStartupPolicy` predicates.
- All 5 CI checks pass on b6ad897, including both reproducible release builds.
- Device pass on a Pixel 9a (Android 16, API 37): start → hosting → stop leaves `mGroup null` with ordering `Group removed successfully` → `P2P channel closed` → `restoring Wi-Fi Aware`. Disabling Wi-Fi mid-hosting exercises both P1 paths in one trace — the second stop logs `Teardown already in progress; chaining completion`, and revalidation logs `Current group 'null' is not ours ('DIRECT-BC-KBEBBQCF'); leaving it alone`, so no device-scoped removal is issued and the Aware lease is released only once teardown resolves.

**One thing worth a decision rather than a change:** the teardown has no timeout, so if the framework never answers `requestGroupInfo`/`removeGroup` — a Wi-Fi HAL or supplicant restart, since the channel is initialised with a `null` `ChannelListener` — `teardownInProgress` stays set on that manager instance. The blast radius is small and self-limiting: `HotspotManager` is constructed per session, so the stuck flag dies with it; the ViewModel's idempotent 10 s fallback still releases the Aware lease; and the orphaned group is recognised by `ownedGroupName` on the next start and removed silently. That leaves one leaked channel registration until the process exits. Same profile as the process-killed-while-hosting case the PR already accepts, so I'd leave it — happy to add a bounded watchdog if you'd rather close it.

PR #833, opened 2026-07-30
Optimize geohash globe rendering and gestures

a1denvalu3:
## Summary

- Improves globe performance on lower-end devices by reducing per-frame calculations and separating static drawing from moving content.
- Stabilizes dragging and flinging with velocity limits, frame-based animation, and multi-touch guards to prevent accidental runaway spins.
- Adds saved **Fast**, **Medium**, and **High** rendering modes. Medium is the default.
- Keeps the quality-selector labels and checkmark aligned when switching modes.
- Fixes incomplete geohash outlines, hidden island coastlines, missing land, and occasional whole-globe land fills.
- Keeps the picker fully native with Jetpack Compose.

## Validation

- `./gradlew :app:testDebugUnitTest :app:assembleDebug`
- `./gradlew :app:lintDebug`
- Visually tested on Pixel 4a (5G) and Pixel 7a in all three rendering modes.

### DIFF
--- a/app/src/main/java/com/bitchat/android/ui/GeohashPickerActivity.kt
+++ b/app/src/main/java/com/bitchat/android/ui/GeohashPickerActivity.kt
@@ -95,13 +95,18 @@ class GeohashPickerActivity : OrientationAwareActivity() {
             BitchatTheme {
                 val context = LocalContext.current
                 val scope = rememberCoroutineScope()
+                val initialRenderQuality = remember(context) {
+                    GlobeRenderQualityPreference.load(context)
+                }
+                var renderQuality by remember { mutableStateOf(initialRenderQuality) }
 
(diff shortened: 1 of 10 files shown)


a1denvalu3:
The renderer sometimes could not tell which side of a large coastline was land.

  For partially hidden continents like Eurasia, it creates an artificial edge along the globe’s horizon
  to close the visible shape. We initially checked only the globe’s center to decide whether that closed
  shape was facing the wrong way.

  That single point was ambiguous in some positions. When it produced the wrong answer, the renderer
  inverted the shape—coloring everything outside the continent as land. Since “outside” covered nearly
  the entire visible globe, the whole globe appeared land-filled.

  The fix checks several points across the globe before deciding to invert a shape, making the decision
  much more reliable.

PR #838, opened 2026-07-31
Stream OpenStreetMap data in geohash globe

callebtc:
## Summary

- replace bundled GeoJSON geography with visible-area OpenStreetMap Shortbread vector tiles
- add bounded disk/memory caching, defensive MVT decoding, tile selection, projection culling, and batched border rendering
- preserve country/state lines, labels, and geohash cells during interaction
- animate newly streamed and removed place labels with a short fade-and-scale transition

## Testing

- `./gradlew --no-daemon --max-workers=2 :app:testDebugUnitTest :app:assembleDebug`
- `./gradlew test :app:lintDebug`
- exercised taps, rotation, zoom, and streamed tile transitions on an Android 17 emulator
- installed and launched on a physical Android device

## Performance

A repeated six-gesture emulator run produced 3.08% jank across 227 frames, with 32 ms median and 34 ms p95 frame time and no missed-vsync frames.


callebtc:
## Android UI visual review

I resolved the comparison as the PR's actual merge-base (`87184edddec2b997dad5c5ffc7d76404c4788c00`) versus the PR head (`033e224bc72d38823aecd1b23c31501771d9e54f`).

### Before / after surface

| Surface | Before | After |
|---|---|---|
| Geohash picker globe | The picker renders bundled world geometry through the local globe data path. | The picker renders streamed OpenStreetMap vector-tile data, adds loading/error-retry feedback and OpenStreetMap attribution, and keeps the existing globe controls and selection flow. |

### Capture limitation

I could not produce or attach reproducible before/after screenshots in this run. The isolated review checkout could not start its debug build because no usable Java Runtime was available; the environment also had no Android SDK/emulator or `adb` available for installation and capture. Therefore this comment is code-level diff analysis only, not screenshot-verified visual evidence. No visual regression or visual improvement is claimed.

The required follow-up is to rerun this review in an environment with JDK 21, the Android SDK, and an emulator, then capture the initial globe, loaded streamed-map state, loading state, and map-error/retry state on both revisions.

Chessing234 (commented):
tile urls encode the viewport (z/x/y). even if this goes through OkHttpProvider/tor, the tile host still sees which cells you pan. either ship a coarse local land layer for the default zoom or say in-ui that opening the globe talks to osm with your view.

PR #847, opened 2026-08-01
Hide the command and mention popups after a message is sent

### LINKED ISSUE #250: CommandSuggestionBox remains after sending the command
### Checklist

### Affected version

0.8.2

1. Write a command without clicking on CommandSuggestionBox
2. Send a command

### Expected behavior

When the command is sent, the CommandSuggestionBox should be hidden.

### Actual behavior

CommandSuggestionBox remains after sending the command.

https://github.com/user-attachments/assets/440f2adf-5e51-43fe-881a-228f56c41571

Android 13

### Affected device model

Redmi 10C

### Additional information

_No response_

heyaim:
# Description

When you type a slash command and send it without tapping the autocomplete list, the command popup stays on screen with its old suggestions. The send path clears the input field in code, and the popup is hidden only from the field's text-change handler, which a code-driven clear does not run, so the flag that shows the popup is never turned off. The mention popup (the `@name` list) has the same cause.

Fixes #250.

## Changes

**Add `CommandProcessor.clearSuggestions()` and call it from `ChatScreen`'s send handler.** The new function hides both popups and empties their suggestion lists, the same resets `selectCommandSuggestion` and `selectMentionSuggestion` already do inline. `ChatViewModel` exposes it next to the other suggestion methods, and the send handler calls it where the field is already cleared.

**Stop the private chat sheet from feeding the shared suggestion state.** That composer called `updateMentionSuggestions` on every keystroke while rendering its own popups as hidden, so the only effect was on the main composer behind it: typing `@` in the sheet and then sending or dismissing left a stale mention popup on the main screen. The sheet keeps its draft handling and drops that call.

## Tests

Two cases in `CommandProcessorTest`:
- `clearSuggestions hides the command suggestion popup` opens the popup with `/`, calls `clearSuggestions()`, and asserts the popup is hidden and its list empty.
- `clearSuggestions hides the mention suggestion popup` does the same for the mention popup, opened with `@a` against a stub peer.

These cover the function the send handler calls. The unit tests here do not drive Compose, so the send handler itself is not exercised.

## Validation

- `./gradlew testDebugUnitTest`: 736 tests, 0 failures, 3 pre-existing skips.
- `./gradlew lintDebug`: passes, no new issues on the changed files.
- `./gradlew assembleDebug`: builds.
- Reproduced on an Android 15 emulator: on main, sending `/w` leaves the popup on screen over the cleared field; with this change the popup is dismissed. Screenshots below.

## Does not do

No change to sending, to what the suggestions contain, or to how the popups open while typing in the main composer. Voice, image, and file sends are untouched.

## Checklist

### DIFF
--- a/app/src/main/java/com/bitchat/android/ui/ChatScreen.kt
+++ b/app/src/main/java/com/bitchat/android/ui/ChatScreen.kt
@@ -386,6 +386,9 @@ fun ChatScreen(viewModel: ChatViewModel) {
                     if (accepted) {
                         messageText = TextFieldValue("")
                         viewModel.setConversationDraft(selectedPrivatePeer, "")
+                        // Clearing the field in code does not run onMessageTextChange,
+                        // so the popups have to be dismissed here.
+                        viewModel.clearSuggestions()
                         forceScrollToBottom = !forceScrollToBottom
                     }
                 }
--- a/app/src/main/java/com/bitchat/android/ui/ChatViewModel.kt
+++ b/app/src/main/java/com/bitchat/android/ui/ChatViewModel.kt
@@ -1341,7 +1341,11 @@ class ChatViewModel(
     fun updateCommandSuggestions(input: String) {
         commandProcessor.updateCommandSuggestions(input)
     }
-    
+
+    fun clearSuggestions() {
+        commandProcessor.clearSuggestions()
+    }
+
     fun selectCommandSuggestion(suggestion: CommandSuggestion): String {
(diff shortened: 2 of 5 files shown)


heyaim:
Before on main, after with this change.
Screenshot 2026-07-31 at 11 30 48 PM
Screenshot 2026-07-31 at 11 30 54 PM

callebtc (approved):

callebtc:
thank you, great fix

PR #854, opened 2026-08-02
Re-sync suggestion popups when the composer text is set in code

heyaim:
# Description

When a suggestion popup is open and the app sets the composer text itself, the popup stays up over the new text. The popups are updated only from the field's text-change handler, which a code-driven change does not run. #847 fixed this for the send path, the case #250 reported. While working on that fix, two more places with the same older bug were found:

- Tapping a nickname in the message list, which inserts `@name ` into the field.
- Switching private conversations, which restores that conversation's draft into the field.

## Changes

**Re-sync both popups with the new text at both sites.** Instead of a blanket clear, each site calls `updateCommandSuggestions` and `updateMentionSuggestions` with the text it just set, the same calls the text-change handler makes while typing. The popup state then always matches the field content: an inserted mention ends with a space, so an open popup hides, and a restored draft that starts with `/` brings the command popup back the way typing it would.

The iOS app already works this way: its composer drives autocomplete from `onChange(of: messageText)`, which runs for code-driven changes too, so a mention insert or a restored draft re-syncs the popup there. This change brings Android to the same behavior.

## Tests

Three cases in `CommandProcessorTest`:
- `text set to an inserted mention hides both popups` pins the nickname-tap shape (`/ @alice `).
- `a mention followed by a space no longer matches the mention popup` pins the trailing-space rule the fix relies on.
- `a restored command draft reopens the command popup` pins the draft-restore behavior for a command draft and for an empty draft.

The unit tests do not drive Compose, so they cover the functions the two sites call rather than the sites themselves.

## Validation

- `./gradlew testDebugUnitTest`: 748 tests, 0 failures, 3 pre-existing skips.
- `./gradlew lintDebug`: passes, no new issues on the changed files.
- `./gradlew assembleDebug`: builds.

Both sites need a second peer to reach from the UI (a nickname on someone else's message, or a private conversation), so I could not walk them on a single emulator the way #847's send path was walked. The behavior each site relies on is pinned by the tests above.

## Does not do

No change to how the popups behave while typing, to sending, or to what the suggestions contain. The send-path dismissal from #847 is untouched.

## Checklist
- [ ] I have mentioned the corresponding issue and the relevant keyword (e.g., "Closes: #xy") in the description (see <https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue>) — no issue exists for these two; found while working on #250 (PR #847)

### DIFF
--- a/app/src/main/java/com/bitchat/android/ui/ChatScreen.kt
+++ b/app/src/main/java/com/bitchat/android/ui/ChatScreen.kt
@@ -100,11 +100,14 @@ fun ChatScreen(viewModel: ChatViewModel) {
     var isScrolledUp by remember { mutableStateOf(false) }
 
     LaunchedEffect(selectedPrivatePeer) {
-        messageText = TextFieldValue(
-            selectedPrivatePeer
-                ?.let(viewModel::conversationDraft)
-                .orEmpty()
-        )
+        val draft = selectedPrivatePeer
+            ?.let(viewModel::conversationDraft)
+            .orEmpty()
+        messageText = TextFieldValue(draft)
+        // Setting the field in code does not run onMessageTextChange, so the
+        // popups have to be re-synced with the restored draft here.
+        viewModel.updateCommandSuggestions(draft)
+        viewModel.updateMentionSuggestions(draft)
     }
 
     // Show password dialog when needed
@@ -342,6 +345,11 @@ fun ChatScreen(viewModel: ChatViewModel) {
                         text = newText,
                         selection = TextRange(newText.length)
                     )
+                    // Setting the field in code does not run onMessageTextChange, so
+                    // the popups have to be re-synced with the new text here. The
+                    // inserted mention ends with a space, which hides an open popup.
+                    viewModel.updateCommandSuggestions(newText)
+                    viewModel.updateMentionSuggestions(newText)
                 },
                 onMessageLongPress = { message ->
                     // Message long press - open user action sheet with message context
--- a/app/src/main/java/com/bitchat/android/ui/ChatScreen.kt
+++ b/app/src/main/java/com/bitchat/android/ui/ChatScreen.kt
@@ -348,6 +348,9 @@ fun ChatScreen(viewModel: ChatViewModel) {
                     // Setting the field in code does not run onMessageTextChange, so
                     // the popups have to be re-synced with the new text here. The
                     // inserted mention ends with a space, which hides an open popup.
+                    // The draft write is skipped the same way; without it, switching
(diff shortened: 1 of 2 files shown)


Chessing234 (commented):
The diagnosis is right and the two sites you fixed are real. But the same bug is still live two handlers below, in the same file.

`onCommandSuggestionClick` and `onMentionSuggestionClick` (`ChatScreen.kt:428-441`) both set `messageText` in code, and neither re-syncs. They aren't fully covered by the selection functions either:

- `selectCommandSuggestion` clears only the **command** popup (`setShowCommandSuggestions(false)`).
- `selectMentionSuggestion` clears only the **mention** popup.

So with `/msg @ali` in the field and both popups open, tapping a mention suggestion produces `/msg @alice ` and hides the mention popup, while the command popup stays up against text that no longer matches it. That's the same "popup stays over the new text" symptom the PR is about. Adding the two `updateXSuggestions` calls at both sites would close it.

Separately, the nickname-tap site you did fix sets `messageText` but never calls `viewModel.setConversationDraft(selectedPrivatePeer, newText)`, unlike `onMessageTextChange` right below it. So a tapped-in mention isn't persisted — switch conversations and back and it's gone, restored from the stale draft by the very `LaunchedEffect` this PR also touches. Worth handling in the same pass.

The three `CommandProcessorTest` cases are good but they test the processor, not the call sites, so they'd pass unchanged with the two sites above still broken. Not asking for Compose tests — just flagging that the coverage doesn't pin what regressed.

heyaim:
Thanks. All three programmatic-edit sites now run the same block a keyboard edit runs: persist the draft, re-sync both popups.

The draft gap is the one that reaches the user. The nickname tap set the field in code and never wrote the draft, unlike the keystroke path in `onMessageTextChange`, and the restore on conversation switch overwrites the field from the stored draft, which never got the mention. The two suggestion handlers skipped the draft write the same way, and all three are fixed together.

On the stale-popup scenario at the two handlers: the calls are added, and I went looking for the both-popups-open state. It does not exist on this branch. Only two calls make either popup visible, one inside each update function. The command popup needs the whole input to prefix a command name or alias (`filterCommands` uses `startsWith` on the full input), the mention popup needs an @, and no command or alias contains an @ or a space. An input that can open one has already closed the other, and each select function closes its own popup on tap. A new `CommandProcessorTest` case pins the closing half. The field is set in code at six places in this screen: five re-sync both popups, and clear-on-send dismisses them, the same result for empty text.

Agreed on coverage: the tests pin the processor, and the call-site wiring stays Compose-level.

PR #857, opened 2026-08-02
UI dynamic size component

spradhan7656:
# Description

## Checklist


spradhan7656:
please add verification.xml my side does not support. if i add the the app can't sync properly

spradhan7656:
i added the dynamic screen wise size component  ssp and spd method the both method help compent size show equal every screen

spradhan7656:
i want to change every screeen to sp and dp to ssp and sdp because some phose i check basically 720p display phone ui break.

callebtc:
## Android UI verification

Captured from clean before/after debug builds on **Android 17 (API 37)** at 360 dp and 411 dp phone widths in light theme. The before build is PR #857's actual merge-base (`a49f3821`); the after build is `459fa27e`.

### Visual changes detected

- At 360 dp, the converted Tor-row geometry remains at its design baseline; only the `RECOMMENDED` badge text shrinks from 10 sp to 9 sp.
- At 411 dp, the Tor title/subtitle and converted spacing scale up with the viewport. The badge stays near its previous apparent size because `9 × 411 / 360 ≈ 10.3 sp`; the row remains contained without clipping.
- The About Info control has no app-content pixel delta. The changed shared `SheetSectionLabel` primitive is not currently rendered there.

### About Info — 411 dp control (no visual delta)

| Before | After |
|---|---|
| Before About Info at 411 dp | After About Info at 411 dp |

### Location Channels Network row — 360 dp

| Before | After |
|---|---|
| Before Location Channels Network row at 360 dp | After Location Channels Network row at 360 dp |

### Location Channels Network row — 411 dp

| Before | After |
|---|---|
| Before Location Channels Network row at 411 dp | After Location Channels Network row at 411 dp |

Fixture disclosure: synthetic nickname only, empty peer state, and BLE/Wi-Fi transports disabled. The app's fail-closed live-location gate remained disabled; no coordinate, geohash, real peer, message, contact, or account data was used. The temporary debug-only fixture was removed after capture.

Limitation: system status-bar time differs between captures; the About app-content region is pixel-identical. Coverage is limited to 360 dp and 411 dp phone widths in light theme.

Chessing234:
the .ssp() helper multiplies sp values by a screen-width scale factor on top of the system font-scale that .sp already applies — that compounds with a user's accessibility text-size setting instead of respecting it, and there's no opt-out. also gradle/verification-metadata.xml lost its trusted-artifacts block (javadoc/sources jars) which has nothing to do with dynamic sizing, looks like an accidental side effect of regenerating metadata locally. can you revert the verification-metadata change and reconsider ssp() so it doesn't stack with system font scaling?

Chessing234 (changes requested):
ssp() scales type off screen width, which fights the user's font-scale setting. also please drop the verification-metadata.xml hunk — removing trusted javadoc/sources artifacts is unrelated to the ui change and will fail gradle for anyone pulling those jars.

PR #859, opened 2026-08-02
docs: flag MAX_PAYLOAD_LENGTH must stay in sync with iOS

harshjainnn:
Companion to permissionlesstech/bitchat#1618, which raised iOS's decompressed payload ceiling from ~1.13 MiB to 10 MiB to match this constant. Adding a comment here so the next change to either constant doesn't silently reintroduce the cross-platform mismatch.

Refs #1618

# Description

This is a documentation-only companion to permissionlesstech/bitchat#1618. That issue was caused by iOS's `FileTransferLimits.maxFramedFileBytes` (~1.13 MiB) and Android's `MAX_PAYLOAD_LENGTH` (10 MiB) silently going out of sync, which caused legitimate large compressed messages from Android to be silently dropped on iOS. The iOS PR raised its ceiling to 10 MiB to match this constant.

This PR adds a comment above `MAX_PAYLOAD_LENGTH` here noting it must stay in sync with iOS's `FileTransferLimits.maxFramedFileBytes`, so the next change to either value doesn't silently reintroduce the mismatch. No value or behavior change.

## Checklist

### DIFF
--- a/app/src/main/java/com/bitchat/android/util/AppConstants.kt
+++ b/app/src/main/java/com/bitchat/android/util/AppConstants.kt
@@ -67,6 +67,9 @@ object AppConstants {
 
     object Protocol {
         const val COMPRESSION_THRESHOLD_BYTES: Int = 100
+        // MUST stay in sync with iOS's FileTransferLimits.maxFramedFileBytes
+        // (LocalPackages/BitFoundation/Sources/BitFoundation/FileTransferLimits.swift in permissionlesstech/bitchat).
+        // Reference issue #1618, which fixed a mismatch where iOS's ceiling was ~1.13 MiB instead of 10 MiB.
         const val MAX_PAYLOAD_LENGTH: Int = 10_485_760
     }
 


Chessing234 (commented):
The intent is good, but the comment as written points at a state that doesn't exist yet, which is the opposite of what a sync-warning comment should do.

Three corrections:

1. **Path casing.** The repo is `localPackages/BitFoundation/...`, lowercase `l`. As written the path doesn't resolve.
2. **#1618 hasn't fixed anything.** #1618 is an *open issue*, not a merged fix, and the iOS PR that raises the ceiling (#1620) is still open. So iOS's `maxFramedFileBytes` is *currently* still ~1.13 MiB. A comment saying it "fixed a mismatch where iOS's ceiling was ~1.13 MiB instead of 10 MiB" tells the next reader the platforms already agree, when the mismatch this is meant to guard against is live right now.
3. **`maxFramedFileBytes` isn't a constant to sync against.** It's computed:

   ```swift
   return maxPayloadBytes + tlvEnvelopeOverhead + binaryEnvelopeOverhead
   ```

   from `maxPayloadBytes` (1 MiB) plus TLV and binary framing overhead. "MUST stay in sync with iOS's `maxFramedFileBytes`" isn't actionable, because nobody edits that number directly — they'd edit `maxPayloadBytes`. And the two values shouldn't be *equal* anyway; the invariant is `MAX_PAYLOAD_LENGTH >= iOS maxFramedFileBytes`, an inequality.

Suggest rewording to state the inequality, name `maxPayloadBytes` as the thing that actually moves, and reference #1620 as pending rather than done. Happy to see this land once the comment describes the real relationship — as-is it would mislead exactly the person it's written for.

Issue #848, opened 2026-08-01
spec review: rotating peer IDs — cross-platform design in permissionlesstech/bitchat#1487


jackjackbits:
iOS just landed the first deliverable for peer ID rotation: a spec plus working, tested derivation primitives — deliberately **not** wired into the shipping mesh. Nothing changes on the air until both platforms agree.

**Spec:** https://github.com/permissionlesstech/bitchat/blob/main/docs/PEER-ID-ROTATION.md (written to be implementable from the document alone — the three test vectors in §7 were reproduced byte-for-byte by an independent Python implementation written only from the text, so you should never need to read the Swift)

**The problem:** the 8-byte peer ID is `SHA-256(noiseStaticKey)[0..8]` and never rotates, and the announce broadcasts both long-term keys plus the nickname in cleartext every few seconds. A passive BLE listener in a crowd gets a permanent per-device identifier and the adjacency graph for free.

**Design in brief:** hour epochs (wall clock, ±1 tolerance) · IDs from `HKDF → HMAC(epoch)` over private key material · pairwise recognition tags from the X25519 shared secret so known peers recognize each other with no handshake (fixed 64-byte tag block so the count doesn't leak) · strangers handshake first, identify second · nickname moves inside the session · neighbor list dropped · a replacement identity binding (Ed25519 proof over context/epoch/rotating-ID/static-key, verified against the pinned key inside the Noise session) because `peerID == hash(staticKey)` is currently the anti-forgery mechanism on both platforms.

**What we'd like from Android:**

1. **Argue with the spec** — every constant and context string is a concrete proposal you can reject by changing one function and watching a test vector move. The open questions where your opinion decides the design:
   - **O1** rotation period — is 1 hour right? shorter is less linkable but churns sessions/routes.
   - **O4** v2 announces are unsigned (no key to verify against without disclosing one). Acceptable if unverified presence stays out of the people list until a tag match or handshake? The alternative is per-epoch ephemeral signing keys + continuity proofs — more machinery, more bytes.
   - **O5** rotate while a Noise session is live, or defer to session close? Deferring is simpler but pins the ID longest for exactly the people who talk most.
   - **O7** does the Android decoder tolerate trailing bytes after the last parsed field? This gates whether extending padding coverage can ship ungated or needs its own capability bit.
[...]

PR #862, opened 2026-08-04
feat(protocol): add peer ID rotation phase 1

a1denvalu3:
## Summary

- implement the peer-ID rotation HKDF/HMAC, X25519 recognition-tag, fixed tag-block, epoch-window, and Ed25519 binding primitives
- add strict `announceV2` (`0x2C`) TLV encoding/decoding and advertise capability bit 14 in existing v1 announces
- parse and discard v2 announces in Phase 1 so unsigned presence cannot establish liveness, enter the people list, or be gossip-relayed
- document Android positions on O1/O4/O5/O7 and pin proposed joint binding-signature and full-packet vectors

## Interoperability vectors

The JVM tests reproduce all three published section 7 values byte-for-byte, including the directional recognition tags. They also pin independently generated joint vectors with fully documented fixed inputs so iOS can reproduce them from the prose.

## Rollout and risk

This is additive Phase 1 only. Android continues to emit v1 announcements and use stable IDs; no rotating ID is used by the shipping mesh. The primary compatibility risk is the new capability advertisement and message-type recognition.

## Validation

- `./gradlew :app:assembleDebug :wear:assembleDebug clientRewriteContractTest`
- focused peer-ID rotation, announce-v2, and capability vector tests
- Mesh Lab: blocked. This PR does not emit v2 on air. A regression handoff is `broadcast`, `dm`, and `raw` after authorized setup.

Refs #848

### DIFF
--- a/app/src/main/java/com/bitchat/android/crypto/PeerIdRotation.kt
+++ b/app/src/main/java/com/bitchat/android/crypto/PeerIdRotation.kt
@@ -1,0 +1,158 @@
+package com.bitchat.android.crypto
+
+import org.bouncycastle.crypto.digests.SHA256Digest
+import org.bouncycastle.crypto.generators.HKDFBytesGenerator
+import org.bouncycastle.crypto.macs.HMac
+import org.bouncycastle.crypto.params.Ed25519PrivateKeyParameters
+import org.bouncycastle.crypto.params.Ed25519PublicKeyParameters
+import org.bouncycastle.crypto.params.HKDFParameters
+import org.bouncycastle.crypto.params.KeyParameter
+import org.bouncycastle.crypto.params.X25519PrivateKeyParameters
+import org.bouncycastle.crypto.params.X25519PublicKeyParameters
+import org.bouncycastle.crypto.signers.Ed25519Signer
+import java.nio.ByteBuffer
+import java.nio.ByteOrder
+import java.security.SecureRandom
+
+/** Cross-platform primitives from docs/PEER-ID-ROTATION-ANDROID.md. */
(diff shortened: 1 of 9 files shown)


Chessing234 (commented):
checked the allocations against the ios side: `announceV2 = 0x2C` and bit 14 both match `docs/PEER-ID-ROTATION.md` §4.6/§5 exactly, including the reasoning for skipping 11–13. good.

the thing i'd push back on is advertising it. `LOCAL_SUPPORTED` now includes `PEER_ID_ROTATION`, but phase 1 parses v2 announces and throws them away. in the ios flow that bit is the signal a peer uses to decide rotation is safe — step 1 of §5 is "A hears an announce from unknown `peerID_e` **advertising the rotation capability**". so an ios build that implements rotation, sees this bit, and starts rotating becomes invisible to this client, because the announces carrying its new IDs are exactly the ones being discarded. that failure is silent and looks like the peer left.

it also cuts against the rule #876 is adding to `ANNOUNCEMENT_GOSSIP.md` in parallel — "implementations advertise only wire features they implement". parsing-and-discarding isn't implementing.

phase 1 reads fine without the advertisement: keep the constant, keep the parser, keep the vectors, and add the bit to `LOCAL_SUPPORTED` in the phase that can actually act on a rotating peer.

smaller: the spec says whoever implements first "pins the number in a shared test vector", and ios's `PeerCapabilities.swift` still has nothing at 11–14, so right now bit 14 exists only in a doc paragraph and in `AnnounceV2Test` here. worth flagging on the ios side so the number is claimed in both claim tables rather than one.

PR #863, opened 2026-08-04
feat: preserve original bytes during re-encoding of packets with foreign encoders

areebahmeddd:


Chessing234 (commented):
The approach is sound and I checked the concern I expected to have: `BitchatPacket` overrides `equals`/`hashCode` manually, so adding `wirePayload` to the primary constructor genuinely doesn't affect packet identity, and your comment is accurate. Tying the bytes to `forPayload` is the right guard.

Two things before this can be reviewed properly:

**The description is empty.** This changes the preimage used for signature verification on a protocol-level path, and it's a port of #1631 on iOS. The reasoning in the `WirePayload` KDoc is good — please put it in the PR body, along with the link to #1631 and the repo's checklist. Right now a reviewer has to reconstruct why this exists from the diff alone.

**The test covers one of the three paths.** `re-encoding preserves a foreign encoder's compressed payload` is a nice construction (the stored-block trick is clever), but it only exercises `wire.compressed == true`. Untested:

- The `forPayload` fallback. The KDoc says "replace the payload and the encoder compresses instead" — that's the safety property the whole design rests on, and nothing asserts it.
- The uncompressed-on-the-wire branch, where the comment says `shouldCompress` agrees across clients but the "only if smaller" check need not. That's the subtler of the two cases and it has no test at all.

One question: `WirePayload` now keeps the compressed bytes alive for the packet's lifetime, on top of the decompressed payload. For packets sitting in dedup caches and relay queues near `MAX_PAYLOAD_LENGTH` that's a real increase in retained memory per packet. Have you measured it, or is there a reason it's bounded that I'm missing? Worth a note in the description either way.

callebtc (approved):

callebtc:
great PR thank you

Issue #351, opened 2025-08-30
Private messages do not work and do not arrive.


marcorizzo97:
Another problem I noticed is that when I send private messages to users who are online on Geoshash, the messages are not delivered to the recipient. There is a small circle next to the message, whereas there should be a tick if it has been received and two ticks if the message has been viewed.
When I test this with various users privately, they cannot see my messages, just as I cannot see theirs, even though they send them to me. It must be a bug.

PR #869, opened 2026-08-11
fix(nostr): cap DM envelope backdating at 24 hours

### LINKED ISSUE #858: Android stamps DM envelopes up to 48h back but iOS only fetches 24h
### Checklist

### Affected version

1.7.4

1. Set up an Android device and an iOS device as mutual favorites, then move them out of Bluetooth range so DMs travel over the internet.
2. Restart the iOS app, so its relay subscription is freshly created.
3. Send ten DMs from the Android device.
4. Count what arrives on the iOS device.

### Expected behavior

All ten arrive.

### Actual behavior

Some of the ten never arrive, and a lost DM never arrives later. The loss rate follows from two constants:

- Android stamps the outer envelope of every DM (the NIP-17 gift wrap) with a random timestamp up to 48 hours in the past. `randomizeTimestampUpToPast` in `NostrCrypto.kt` (line 322) defaults to 172800 seconds and draws uniformly. `createSeal` and `createGiftWrap` in `NostrProtocol.kt` both use it. This follows NIP-17, which says clients "SHOULD randomize `created_at` in up to two days in the past in both the seal and the gift wrap".
- iOS asks relays only for gift wraps stamped within 24 hours before the subscription was created: `nostrDMSubscribeLookbackSeconds = 86400` in its `TransportConfig.swift`, used by every DM subscription. There is no wider catch-up query and no saved position between launches.

A gift wrap stamped behind the `since` value of the iOS subscription fails the relay filter. Relays apply `since` to stored events and to events that arrive while a subscription is open (NIP-01), and being online at the same time does not help. I verified this against the four default relays. relay.damus.io, relay.primal.net and offchain.pub each accepted a test gift wrap stamped 30 hours in the past, returned it to a subscription asking 48 hours back, and never returned or pushed it to one asking 24 hours back. The fourth, nostr21.com, rejects gift wraps outright ("blocked kind 1059 reason: this is not a DM relay") and cannot deliver DMs at all. In a second run against relay.primal.net, 20 gift wraps stamped with the same uniform draw the Android code uses were delivered or lost by a 24 hour subscription exactly along the boundary: all 7 stamped under 24 hours arrived, all 13 stamped older did not.

On two phones, that math says about five of the ten arrive. Which ones is random per send. The number comes from reading the code on both sides and from the relay runs above; I do not have a second phone to run the two phone test on.

Half is the number for a freshly launched app. iOS reuses the relay request it built at launch, which accepts anything stamped in the 24 hours before that launch or later (`NostrRelayManager.flushPendingSubscriptions`). The longer the app runs, the older that boundary gets, and the more backdated stamps clear it. A restart sets a new boundary and the loss returns to half. Geohash DMs sit near half regardless. A missed DM never arrives later: every new boundary is later than the one it missed.

The same envelope path carries more than messages. Delivery acks, read receipts, and favorite notifications all go through `NostrProtocol.createPrivateMessage` and get the same randomized stamp (`NostrTransport.kt`: `sendDeliveryAck`, `sendNextReadAck`, `sendFavoriteNotification`, and the geohash variants). So the damage shows on both phones: a message from Android to iOS is lost outright half the time, and when a message from iOS to Android arrives fine, the DELIVERED and READ receipts Android sends back are lost at the same rate, leaving the iOS sender stuck on the sent state for a message that actually arrived.

Each app is consistent with itself. Android's own DM subscription looks back 48 hours, with the comment "align with NIP-17 randomization" (`NostrClient.kt` line 146, in `subscribeToPrivateMessages`). iOS randomizes its own envelopes by only plus or minus 15 minutes (`randomizedTimestamp` in its `NostrProtocol.swift`) and looks back 24 hours. Android to Android and iOS to iOS lose nothing. Message content is only lost from Android to iOS. Delivery indicators can look broken in both directions, because Android's receipts to an iOS sender travel the same Android to iOS path.

Not applicable. The loss happens at the relay filter, before anything reaches the screen.

Any. The cause is in the sending code and does not depend on the OS version.

### Affected device model

Any.

### Additional information

DMs over the Bluetooth mesh are unaffected. This covers DMs that travel over Nostr: mutual favorites out of Bluetooth range, and geohash DMs.

The 48 hour randomization arrived in PR #305, merged Aug 24, 2025, first shipped in release 1.2.0 on Aug 25, 2025. The PR description notes "ios to android works" and "android to ios still wip". Three open reports filed after that release match what this bug would look like from the outside: #351 (geohash private messages not delivered, no ticks), #391 (DMs stuck on the sent circle, and the reporter says it used to work), and #458 (on 1.3.1, private messages between Android and iOS need several resends before one lands). I cannot confirm those reporters' setups from the reports.

Either repo can close the gap. Android could shrink its randomization window to fit inside the 24 hours iOS asks for, which stays within the NIP-17 wording ("up to two days") but gives up some timestamp privacy. Or iOS could widen its lookback to 48 hours, which costs more relay traffic on every launch and is a change in the iOS repo. Which window should move is a design call.

It may be worth a line in the protocol docs: the receiving side's lookback has to be at least as long as the sending side's randomization window.

Once a direction is picked, I can build the fix on either side.

qutad:
## Summary

- Cap Android's outbound NIP-17 timestamp randomization at 24 hours.
- Apply the limit to both gift wraps and encrypted seals.
- Keep Android's 48-hour inbound lookback unchanged.
- Add regression coverage for both envelope timestamps.
- Document the relationship between sender randomization and receiver lookback.

## Problem

Android randomized NIP-17 gift-wrap timestamps up to 48 hours into the past, while iOS subscribes with a 24-hour lookback.

Relays filter events using the gift wrap's `created_at`. As a result, approximately half of newly sent Android envelopes could fall behind a fresh iOS subscription and never be delivered. This also affected delivery acknowledgements, read receipts, and favorite notifications using the same envelope path.

## Solution

Reduce Android's outbound randomization window from 48 hours to 24 hours. This remains valid under NIP-17's "up to two days" recommendation and aligns generated envelopes with the iOS subscription window.
Android continues accepting envelopes up to 48 hours old so interoperability with clients using the wider NIP-17 window is preserved.

## Testing

- `./gradlew :app:testDebugUnitTest --tests com.bitchat.android.nostr.NostrProtocolTest`
- `./gradlew clientRewriteContractTest`
- `git diff --check`

All tests passed.

Fixes #858

### DIFF
--- a/app/src/main/java/com/bitchat/android/nostr/NostrCrypto.kt
+++ b/app/src/main/java/com/bitchat/android/nostr/NostrCrypto.kt
@@ -21,7 +21,9 @@ import java.math.BigInteger
  * Includes secp256k1 operations, ECDH, and NIP-44 encryption
  */
 object NostrCrypto {
-    
+
+    internal const val NIP17_DEFAULT_MAX_PAST_SECONDS = 85_500
+
     private val secureRandom = SecureRandom()
     // NIP-44 v2 only
     
@@ -317,9 +319,9 @@ object NostrCrypto {
     }
     
     /**
-     * Random timestamp up to maxPastSeconds in the past (default 24 hours for iOS compatibility)
+     * Random timestamp in the past, defaulting to 23h45m to leave slack inside iOS's 24h lookback.
      */
-    fun randomizeTimestampUpToPast(maxPastSeconds: Int = 86400): Int {
+    fun randomizeTimestampUpToPast(maxPastSeconds: Int = NIP17_DEFAULT_MAX_PAST_SECONDS): Int {
         val now = (System.currentTimeMillis() / 1000).toInt()
         val offset = if (maxPastSeconds > 0) secureRandom.nextInt(maxPastSeconds + 1) else 0
         return now - offset
--- a/app/src/main/java/com/bitchat/android/nostr/NostrCrypto.kt
+++ b/app/src/main/java/com/bitchat/android/nostr/NostrCrypto.kt
@@ -22,7 +22,7 @@ import java.math.BigInteger
  */
 object NostrCrypto {
 
-    internal const val NIP17_DEFAULT_MAX_PAST_SECONDS = 85_500
+    internal const val NIP17_DEFAULT_MAX_PAST_SECONDS = 79_200
 
     private val secureRandom = SecureRandom()
     // NIP-44 v2 only
@@ -319,7 +319,7 @@ object NostrCrypto {
     }
     
     /**
-     * Random timestamp in the past, defaulting to 23h45m to leave slack inside iOS's 24h lookback.
+     * Random timestamp in the past, defaulting to 22 hours to leave 2 hours of slack inside iOS's 24-hour lookback.
      */
     fun randomizeTimestampUpToPast(maxPastSeconds: Int = NIP17_DEFAULT_MAX_PAST_SECONDS): Int {
         val now = (System.currentTimeMillis() / 1000).toInt()
--- a/app/src/main/java/com/bitchat/android/nostr/NostrCrypto.kt
+++ b/app/src/main/java/com/bitchat/android/nostr/NostrCrypto.kt
@@ -317,9 +317,9 @@ object NostrCrypto {
     }
     
     /**
-     * Random timestamp up to maxPastSeconds in the past (default 2 days)
+     * Random timestamp up to maxPastSeconds in the past (default 24 hours for iOS compatibility)
      */
-    fun randomizeTimestampUpToPast(maxPastSeconds: Int = 172800): Int {
+    fun randomizeTimestampUpToPast(maxPastSeconds: Int = 86400): Int {
         val now = (System.currentTimeMillis() / 1000).toInt()
         val offset = if (maxPastSeconds > 0) secureRandom.nextInt(maxPastSeconds + 1) else 0
         return now - offset
--- a/app/src/main/java/com/bitchat/android/nostr/NostrProtocol.kt
+++ b/app/src/main/java/com/bitchat/android/nostr/NostrProtocol.kt
@@ -37,7 +37,7 @@ object NostrProtocol {
         val rumorId = rumorBase.computeEventIdHex()
         val rumor = rumorBase.copy(id = rumorId)
         
-        // 2. Seal the rumor with 15 minutes of slack inside iOS's 24-hour lookback.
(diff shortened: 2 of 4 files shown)


heyaim:
This is the failure I measured in [#858](https://github.com/permissionlesstech/bitchat-android/issues/858): the gift wrap stamp falls behind the receiver's since filter, and capping the backdate inside the 24 hour lookback fixes it. The open question in that issue was which window should move, Android's randomization or iOS's lookback. Merging this makes the pick by default.

The 15 minutes of slack looks thin to me. It is also what has to absorb any clock difference between the two phones: when the sender's clock runs behind the receiver's, delivery needs the backdate to stay under 86400 minus that difference, and with NIP17_DEFAULT_MAX_PAST_SECONDS at 85500 that leaves 900 seconds. A phone with automatic time turned off can be off by more than 15 minutes. I would reserve an hour or two. It just means backdating a little less.

Chessing234 (commented):
shrinking the window is the right call of the two the issue offered, and 22h is a defensible landing spot. one thing the body overstates though.

the 2h slack only covers the case where the ios peer is already subscribed, or launches within 2h of the send. its `since` is anchored to launch (`T - 86400`), so for a message sent Δ before ios next launches, delivery needs `offset <= 86400 - Δ`. with offset uniform over [0, 79200]:

- Δ = 0 (ios already running): 0% lost — this is the repro in #858, and it's genuinely fixed
- Δ = 6h: ~27% lost
- Δ = 12h: ~45% lost
- Δ ≥ 24h: everything lost, but that's the lookback itself, not the randomization

so store-and-forward to a phone that's been off for a few hours still drops a large slice, and with offset=0 none of those would drop. worth saying in the body that this closes the live-subscription case rather than "approximately half could fall behind" being fixed outright, because the offline case is the one people file bugs about (#351/#391/#458 all read like that).

separately: `NIP17_DEFAULT_MAX_PAST_SECONDS` bakes an ios constant into this repo, and the test asserting `86400 - 7200 == 79200` restates the arithmetic rather than checking anything ios-side, so it can't notice `nostrDMSubscribeLookbackSeconds` moving. same class as #859's MAX_PAYLOAD_LENGTH note — a comment naming the ios symbol would at least make it greppable from the other side.

the ack/receipt path is the part i'd consider splitting out: those carry no timestamp-privacy value worth the delivery cost, and they're why the sender sits on "sent" for a message that actually arrived.

callebtc:
thank you!

PR #872, opened 2026-08-11
Fix reproducible R8 mapping output

callebtc:
## Summary
- enable R8 deterministic-debugging mode for canonical release builds
- stabilize the ProGuard mapping metadata embedded in phone and Wear app bundles
- document the Docker memory requirement for local reproduction

## Context
The v2.0.0 release gate built matching APK payloads but the independently produced AABs differed in their embedded ProGuard mapping metadata. The release was stopped before signing or publication.

## Validation
- shell syntax check passes
- whitespace check passes
- full two-replica reproducibility validation is delegated to this PR CI because deterministic R8 exceeds an 8 GiB Docker allocation locally

## Risk
Release tooling only. The deterministic R8 mode uses one compiler thread and disables randomized input traversal; release builds take longer and require more memory.


habibhabibow77-dev (commented):
q

PR #876, opened 2026-08-13
feat: Add announce capability bit assignments

### LINKED ISSUE #761: Feature: Capability negotiation bits on announce TLV
## Summary
Port iOS **capability negotiation bits** on announce TLVs so peers can discover support for groups, bridge, prekeys, vouch, etc. without guessing.

## Upstream (iOS)
- https://github.com/permissionlesstech/bitchat/pull/1375 — Capability negotiation bits on announce TLV
- Follow-on protocol for private media status: https://github.com/permissionlesstech/bitchat/pull/1434 (authenticated peer-state inner type `0x21`)

## What it is
Announce TLV advertisements of feature support. Capability bits gate:
- Private groups
- Gateway/bridge
- Prekey bundles
- Vouch / transitive verification
- Other optional mesh features

Public capability bits are **discovery only** — they must not alone pin trust or select encrypted media paths; iOS later bound true peer state inside Noise (`0x21`).

## Why Android needs it
Foundation for almost every recent iOS optional feature. Without capability bits, Android cannot advertise or correctly negotiate interop features.

## Acceptance criteria
- [ ] Parse unknown capability TLVs safely (forward compatible)
- [ ] Advertise bits for features Android implements
- [ ] Do not trust discover bits for crypto policy decisions
- [ ] Align with iOS bit assignments
- [ ] Optional: Noise-authenticated peer-state (`0x21`) when shipping encrypted private media (# related)

## Priority
High — prerequisite for groups/bridge/prekeys/vouch interop.

qutad:
## Summary

- Add the iOS-compatible capability bit assignments for prekeys, Wi-Fi bulk, gateway, groups, board, vouch, mesh diagnostics, bridge, private media receipts, and reserved Noise replacement
- Continue advertising only features currently implemented by Android
- Add tests covering bit assignments, minimal little-endian encoding, and forward-compatible decoding
- Document the capability assignments and clarify that advertised capabilities are discovery hints, not trusted crypto policy

## Testing

- [x] `git diff --check`
- [x] `./gradlew testDebugUnitTest lintDebug`
- [ ] Physical two-device Mesh Lab

## Risk

Low. This adds named assignments for existing capability encoding without advertising unsupported features. `LOCAL_SUPPORTED` remains limited to private media.

Closes #761


Chessing234 (commented):
checked all eleven against ios `PeerCapabilities.swift` — 0 prekeys through 10 nonDestructiveNoiseReplacement all line up, and the `0x8000000000000001` round-trip is the right thing to pin.

one thing in the doc bullet. "unknown bits and bytes beyond the low 64 bits must be tolerated for forward compatibility" is true for the announce `0x05` TLV it sits under, but it is not true of the same field in the session-bound form: `AuthenticatedPeerState.decode` here rejects `length !in 1..8`, and ios's `AuthenticatedPeerStatePacket.decode` rejects `value.count > 8` the same way. so a 9-byte capabilities field is truncated-and-tolerated in an announce and drops the whole payload — signing key included — inside 0x21. someone implementing from this paragraph would reasonably write the tolerant version in both places.

related, since you're naming bit 9: ios only honours `privateMediaReceipts` when it arrives in the authenticated 0x21 state (`receiptSessionGeneration` requires bits 8 and 9 off `authenticatedStates`), never off the announce. worth a word, because "advertise only what you implement" reads as if the announce is where the bit does its work, and for 9 it isn't.

PR #877, opened 2026-08-13
feat: Add expanded store-and-forward delivery

### LINKED ISSUE #760: Feature: Expanded store-and-forward (open couriers, durable outbox, 6h public history)
## Summary
Port iOS **expanded store-and-forward** so DMs and public history eventually deliver through crowds of strangers — not only mutual favorites currently connected.

Android has a basic `StoreForwardManager`; iOS substantially extended the model.

## Upstream (iOS)
- https://github.com/permissionlesstech/bitchat/pull/1372 — Expand S&F: open couriers, spray-and-wait, persistent outbox, 6h public history
- https://github.com/permissionlesstech/bitchat/pull/1415 — DMs to unreachable peers use S&F instead of failing instantly
- https://github.com/permissionlesstech/bitchat/pull/1368 — Courier deposit in parallel when only route is a send round
- Bridge integration: https://github.com/permissionlesstech/bitchat/pull/1412

## What it is
- **MessageOutboxStore**: queued private messages persist to disk (sealed); retry across app restarts
- **Open couriers**: strangers can carry sealed mail (quota-bounded), not only mutual favorites at send time
- **Spray-and-wait**: multi-carrier deposit strategy
- **6h public history** persistence for peer serve + echoes UX
- **No instant fail** when peer is neither mesh-reachable nor Nostr-mutual — hand to router/outbox/courier/bridge

## Why Android needs it
Delivery parity is fundamental; without it Android DMs still die when the recipient briefly leaves range, while iOS queues them.

## Acceptance criteria
- [ ] Durable encrypted outbox survives process death
- [ ] Open-courier deposit/pickup wire-compatible with iOS
- [ ] Unreachable DM path never marks failed without enqueue
- [ ] Public history retention window parity (~6h) for sync/echoes
- [ ] Quotas, expiry, panic-wipe behavior

## Priority
High — core messaging reliability; prerequisite for bridge/courier excellence.

qutad:
## Summary
Ports the expanded store-and-forward behavior from iOS so private messages can remain queued and eventually deliver when the recipient is unreachable.

## Changes

- Add a Keystore-sealed persistent DM outbox that survives process death.
- Queue unreachable DMs instead of failing or marking them sent immediately.
- Add iOS-compatible courier envelopes, rotating recipient tags, and Noise X sealing.
- Permit cryptographically verified peers to act as quota-bounded open couriers.
- Add spray-and-wait delivery with persistent copy budgets and per-courier history.
- Require authenticated direct ingress for courier deposits.
- Remove custody or commit spray budgets only after all transport fragments are accepted.
- Add sealed delivery receipts through couriers and Nostr kind-1401 drops.
- Confirm kind-1401 storage through relay `OK` responses before marking bridge deposit complete.
- Persist public messages for six-hour gossip synchronization.
- Retain fragments and public files for the shorter 15-minute synchronization window.
- Add a durable `Queued` delivery state.
- Extend panic wipe to clear outbox, courier storage, history, keys, and pending relay callbacks.
- Add fixed courier wire vectors and persistence, quota, expiry, ingress, fragmentation, and history tests.

## Security and Privacy

- Courier contents remain opaque and Noise X sealed.
- Open courier deposits require a signed, cryptographically verified direct peer.
- Favorite and verified courier tiers have separate quotas and eviction behavior.
- Outbox, courier custody, and public history are bounded and expire automatically.
- Panic wipe independently clears persistent store-and-forward state.
- No identities, message contents, or peer metadata are added to metrics or external services.

## Tests

Passed:
- `./gradlew :app:testDebugUnitTest`
- `./gradlew :app:assembleDebug`
- `./gradlew :app:lintDebug`
- `./gradlew clientRewriteContractTest`
- `git diff --check`

## Validation Gaps

- Physical Mesh Lab validation was not run because two authorized disposable Android devices were unavailable.
- Live Android/iOS courier interoperability was not run.
- The lint task passes but still reports the repository's existing lint findings.

Closes #760


callebtc:
Thank you for the PR! Review:

Re-review at PR head `d83fb82a`, compared against current iOS `main`.

### Blocking findings

1. **[P1] Preserve the iOS prekey discriminator when forwarding courier mail.** `CourierEnvelope.decode()` accepts unknown TLVs but does not retain TLV `0x05`, and `CourierStore.deposit()` stores a newly encoded envelope. A current iOS v2 envelope therefore loses its `prekeyID` while passing through an Android courier; the iOS recipient then attempts the static-key open and drops otherwise valid mail. An isolated round-trip regression confirmed that the seven-byte TLV is removed. Preserve TLV `0x05` (or unknown TLVs/raw encoding) across decode/store/encode.

2. **[P1] Reserve spray budget before starting the asynchronous send.** `CourierStore.sprayCopiesFor()` calculates an offer without mutating or reserving custody, while `BluetoothMeshService` calls `commitSpray()` only after transport completion. A second announce can therefore obtain the same half-budget before the first send completes, including a repeated announce from the same peer. With a budget of four, two previews both offered two copies; after the packets leave, one commit fails but six effective copies exist. Add an in-flight per-envelope reservation and roll it back when transport fails or is cancelled.

3. **[P1] Bound sync responses to the filter's represented type/time window.** Android retains capacity-sized message and fragment collections but combines them into one capacity-sized GCS filter. The responder then checks all collections, so entries omitted only because the filter was full look missing and are retransmitted every 30 seconds. An identical-history regression produced two response packets. Android also ignores the current iOS `types` and `sinceTimestamp` REQUEST_SYNC TLVs. Add type-scoped rounds plus the oldest-covered timestamp cursor, and honor those fields when responding.

### Additional findings

- **[P2] Persist bridge-drop receive dedup across process restarts.** Existing service/global LRUs handle same-process duplicates, but restart clears them. Durable message admission suppresses duplicate UI delivery, while the courier handler still reports success and publishes another acknowledgement. Current iOS persists seen drop event IDs and deduplicates the inner message ID before acknowledgement.
- **[P2] Bound confirmed relay-send handlers.** Acceptance handlers are removed only by relay `OK` or panic cleanup. A silent relay leaves the handler indefinitely after socket-write success, and bridge retries create fresh event IDs and handlers. Resolve failed writes/rejections and add a timeout, as current iOS does.
- **[P3] Same-ID explicit resend can re-use courier slots.** This behavior is reproducible, but current Android UI sends create new IDs and scheduled outbox retries use the stored entry, so I do not consider it merge-blocking without a current resend path.

PR CI and existing focused tests pass, but they do not cover these adversarial cases. Physical Mesh Lab and live Android/iOS courier interoperability were not run for this review.

Chessing234 (commented):
the bridge courier subscription passes milliseconds to `NostrFilter.Builder.since`, which divides by 1000 internally. `now - max_lifetime_ms` therefore becomes a seconds value about 1,000x too far in the future, so relays return no queued drops. pass the millisecond timestamp (or use the builder's conversion once) and add a filter-unit test.

Chessing234 (commented):
the courier subscribe still does `.since(now - CourierEnvelope.MAX_LIFETIME_MS)` and `NostrFilter.Builder.since` divides by 1000. that still sends a seconds value ~1000× too far ahead, so relays return nothing. pass ms into the builder (or convert once) and cover it with a filter unit test — same blocker as before; tip still has the line.

PR #884, opened 2026-08-15
Navigation 3, dependency injection, and module structure for the phone client

moehamade:
# Description

**Draft — opened early for direction, not for merge.** It is one branch so the pieces can be read together; if the direction is agreed I will split it into the smaller PRs listed at the bottom.

Introduces dependency injection, Gradle convention plugins, a first shared module, and a Navigation 3 skeleton to the phone client. Design doc is in `docs/navigation3-modularization-plan.md`, including the parts of it that turned out to be wrong.

The nine chat overlays are deliberately **not** converted yet. That is the large, risky part and it belongs in its own reviewable steps.

## Why

The phone client has no navigation layer (`navigation-compose` is declared but unused — nothing calls `NavHost` or `rememberNavController`), no DI, and 273 files in one module. Navigation today is a `when (onboardingState)` over 8 states, nine overlays driven by independent booleans, and a hand-written back priority chain.

## What is here

- **Hilt + KSP**, with the four ViewModels that are actually obtained from a `ViewModelProvider` injected, and the hand-rolled `ViewModelProvider.Factory` deleted.
- **`build-logic`** convention plugins (`bitchat.android.library`, `.library.compose`, `.hilt`).
- **`:core:domain`** — the two model files the UI actually imports. A library module cannot depend on an application module, so these have to leave `:app` before any feature module can compile.
- **`:core:navigation`** — `Navigator`, an `@ActivityRetainedScoped` back stack, the `EntryProviderInstaller` multibinding, and the `NavDisplay` host.
- **Onboarding and chat as routes**, replacing the `when` and the `OnBackPressedCallback` that was being registered inside a composable body on every recomposition.

## Three things worth a maintainer's eye

**1. Reproducibility.** Adding KSP means codegen inside the byte-for-byte pipeline. Two `:app:bundleRelease` runs under the flags from `build-release.sh` produced identical bytes locally, so the processors do not introduce ordering nondeterminism. That is a proxy, not the guarantee — the canonical check pins JDK 21.0.11+10 and Linux SDK archives and only runs in the container, so CI is the real test. Verification metadata gains components in three reviewed batches and removes none.

**2. `:wear`.** `syncSharedAppSources` gains a second `from(...)` block reusing the same include and exclude lists, so the files moved into `:core:domain` keep being mirrored. `AGENTS.md` is updated to match. Verified `:wear`'s `sharedSrc` still resolves all nine model files and still excludes `FileSharingManager`.

This also caps the DI work: 10 of the stateful singletons are compiled into `:wear`, which has no Hilt and no KSP, so they can be *provided* into the graph but never converted to `@Inject constructor` classes.

**3. `gradle.properties` metaspace 512m → 1g.** KSP and the Dagger processors exhaust the old ceiling. The failure does not name its cause — it surfaces as `ClassLoader.defineClass` OOM in an unrelated task, and if it lands during `--write-verification-metadata` it can truncate `gradle/verification-metadata.xml` to zero bytes. JVM args are not build inputs, so this does not affect reproducibility.

## Two design calls that differ from the doc

**Onboarding is one destination, not eight.** Its steps are driven by permission results and adapter state changes, not by the user navigating, and going back a step has never been supported. Per-step routes would invent a history that does not exist. Only the crossing into chat is a navigation event, and it uses `resetTo` so Back cannot re-enter onboarding.

**`MeshService` stays in `:app`.** It looked like a cheap extraction — it is already an interface — but it returns `noise.NoiseSession.NoiseSessionState` and references three types from its own package, so moving it would pull the Noise/crypto layer into `:core:domain`. Feature modules should not depend on the transport API anyway.

## Verification

`:app:assembleDebug`, `:app:assembleRelease` (R8), `:wear:assembleDebug`, 591 unit tests, and manual checks on an emulator: launch renders onboarding, granting permissions advances without touching the back stack, chat is reached, Back from chat exits rather than re-entering onboarding, and rotation preserves the stack.

The emulator pass was not optional — seeding the back stack from a `LaunchedEffect` compiled and passed every test while crashing on launch, because `NavDisplay` rejects an empty back stack and effects do not run until after the first composition.

## If the direction is agreed, this splits into

1. `build`: KSP + Hilt, lock state and verification metadata
2. `build`: metaspace
3. `feat(di)`: inject the ViewModels
4. `refactor`: convention plugins + `:core:domain` + the `:wear` sync change
5. `feat(nav)`: `:core:navigation`
6. `feat(nav)`: onboarding and chat as routes
7. `refactor(di)`: bridge the singletons that injectable code reached statically

## Checklist

- [x] I have read the contribution guidelines
- [ ] I have mentioned the corresponding issue — **no issue yet.** Opened as a draft to agree direction first; happy to open a tracking issue if you would rather have one.
- [x] If it is a core feature, I have added automated tests — no behaviour is added, and the existing 591 tests cover the refactored paths. Back-order tests are owed when the overlays become routes.


moehamade:
Closing this in favour of smaller, separately reviewable PRs.

This was opened as a direction proposal rather than a merge candidate, and that question is settled — so it has served its purpose. The work is being split rather than dropped:

- #912 — the `OnBackPressedCallback` registered inside a composable body, fixed standalone against the current architecture
- #913 — Navigation 3: Hilt/KSP, injected ViewModels, and onboarding/chat as routes, with the navigation types kept in `:app`

Still to come, once those land:

- Gradle convention plugins, `:core:domain`, moving the navigation types into `:core:navigation`, and the `:wear` sync change
- The remaining DI cleanup (singleton bridging, `MessageRouter` provider, preferences injection)

The branch stays put, so nothing here is lost.

PR #885, opened 2026-08-16
fix: bound the work and traffic one peer can trigger through REQUEST_SYNC

Chessing234:
## What

`REQUEST_SYNC` is handled in `PacketProcessor`'s public/unauthenticated branch (`PacketProcessor.kt:142`) and handed straight to `GossipSyncManager.handleRequestSync`. Nothing on that path limits how often a neighbor may send one.

So a single peer decides how often we:
- decode its GCS filter,
- run SHA-256 over every stored packet (`PacketIdUtil.computeIdBytes`, once per stored packet per request),
- and transmit everything its filter claims to be missing.

A filter that claims to hold nothing is a few bytes on the wire (`p=1, m=1, data=[]`) and makes us replay up to `seenCapacity()` packets — 500 by default. Repeating it costs the sender nothing, and the reply goes out over a link measured in kbps.

## Fix

A token bucket per requester, checked *before* any decoding or hashing:

- **request rate** — one serviced request per 5s, burst 2, so the scheduled first sync (`scheduleInitialSyncToPeer`, 5s) and a periodic one (30s) both land.
- **response allowance** — refills to `seenCapacity()` over the 30s periodic interval, which is the most a peer can legitimately be missing in one round.

An honest peer never reaches either limit, so what it receives is unchanged.

The table is keyed by a peer ID read off the wire, so it's an access-ordered `LinkedHashMap` capped at 256 entries — otherwise the mitigation would be its own memory-exhaustion vector.

## Evidence

`GossipSyncRequestBudgetTest`, 8 tests. Four pin the limits, four pin that normal sync is untouched.

Teeth check — with `admitRequest` bypassed, exactly the four limit tests fail:

```
GossipSyncRequestBudgetTest > requester bookkeeping stays bounded when peer ids are spoofed FAILED
GossipSyncRequestBudgetTest > a backwards clock jump does not hand out free budget FAILED
GossipSyncRequestBudgetTest > back to back requests stop being serviced after the burst allowance FAILED
GossipSyncRequestBudgetTest > a flood cannot pull more packets than the refill window allows FAILED
8 tests completed, 4 failed
```

The other four are the honest-peer guards and pass either way, which is the point of them.

Ran locally with JDK 21:

```
BUILD SUCCESSFUL
```

Measured the same way on `main` and on this branch, the only difference is the new class: **90 classes / 591 tests → 91 / 599**. Lint is unchanged from `main` (305 errors, 333 warnings, 17 hints — all pre-existing/baselined).

Not run locally: instrumented tests and any on-device check. I don't have two radios to hold a real mesh flood against, so the throttle is verified against the manager's own interface rather than over BLE.

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
@@ -654,7 +654,7 @@ class BluetoothMeshService(private val context: Context) : TransportBridgeServic
                 // Decode request and respond with missing packets
                 val fromPeer = routed.peerID ?: return
                 val req = RequestSyncPacket.decode(routed.packet.payload) ?: return
-                gossipSyncManager.handleRequestSync(fromPeer, req)
+                gossipSyncManager.handleRequestSync(fromPeer, req, routed.ingressLinkID)
             }
         }
         
--- a/app/src/main/java/com/bitchat/android/mesh/MeshCore.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/MeshCore.kt
@@ -525,7 +525,7 @@ class MeshCore(
             override fun handleRequestSync(routed: RoutedPacket) {
                 val fromPeer = routed.peerID ?: return
                 val req = RequestSyncPacket.decode(routed.packet.payload) ?: return
-                gossipSyncManager.handleRequestSync(fromPeer, req)
+                gossipSyncManager.handleRequestSync(fromPeer, req, routed.ingressLinkID)
             }
         }
     }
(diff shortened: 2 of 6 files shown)


Chessing234:
good catch — REQUEST_SYNC isn't in the set verifyPacketSignature authenticates, so the sender id is free to rotate and the budget was bypassable exactly as described. rekeyed it on routed.ingressLinkID, which the transport assigns and never puts on the wire. two tests cover it: rotating ids on one link now shares one bucket, and separate links stay independent.

Chessing234:
also added the responder budget to docs/sync.md, since AGENTS.md asks for the spec to move with protocol changes and that file already carries the other receiver-side limits (filter cap, malformed-payload rejection) but nothing about servicing rate. wrote down the sender-id caveat there too — it applies to any implementation, not just this one.

flagging the gate i can't meet: this touches packets and routing, so per AGENTS.md it wants mesh lab validation on physical devices. i don't have two, so the evidence here is the unit suite plus testDebugUnitTest/lintDebug locally. happy to rework if a lab run shows the 5s/30s budgets are too tight for a real multi-peer catch-up.

Issue #610, opened 2026-01-15
Graphene: Tor icon is green (thinks its connected) even if Wifi doesn't work


callebtc:
### Checklist

- [x] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/permissionlesstech/bitchat-android/releases/latest).
- [x] I made sure that there are *no existing issues* - [open](https://github.com/permissionlesstech/bitchat-android/issues) or [closed](https://github.com/permissionlesstech/bitchat-android/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] This issue contains only one bug.
- [x] I have read and understood the [contribution guidelines](https://github.com/permissionlesstech/bitchat-android/blob/main/README.md#contributing).

### Affected version

1.6.0

### Steps to reproduce the bug

the debug settings log indicate that the tor client knows that it's erroring "coul

### Expected behavior

_No response_

### Actual behavior

_No response_

### Screenshots/Screen recordings

```
[...]

PR #886, opened 2026-08-16
fix: stop showing Tor as connected after its circuits stop working

Chessing234:
Refs #610.

## What

Arti's bootstrap milestones only fire on the way up. `handleArtiLogLine` latches the session on `"We have found that guard [scrubbed] is usable."`:

```kotlin
_statusFlow.update { it.copy(state = TorState.RUNNING, bootstrapPercent = 100, running = true) }
stopInactivityMonitoring()
```

After that, nothing downgrades the status except a deliberate stop or the `"Another process has the lock on our state files"` line. There is no post-bootstrap health check at all — `armBootstrapInactivityWatchdog()` returns immediately once `bootstrapPercent >= 100`.

So when the network goes away after bootstrap, exactly what #610 reports happens: Arti logs a continuous stream of

```
ERROR: Failed to connect through Tor: Error { detail: ObtainExitCircuit { ... ChanTimeout ... } }
ERROR: SOCKS connection error: tor: error connecting to Tor: Failed to obtain exit circuit for ports [scrubbed]
```

while the header keeps showing the connected colour — `ChatHeader.kt:169` gates it on `running && bootstrapPercent >= 100`. `isProxyEnabled()` agrees, so `awaitSelectedRoute` returns straight away and requests keep being routed into a SOCKS port that cannot reach an exit.

## Fix

Track circuit failures after bootstrap and drop the published status below 100% once they both accumulate **and** persist.

Circuit failures are a normal part of Tor, so a burst inside one instant is ignored — the report shows three threads failing in the same millisecond. It takes 4 failures that are still going 20 seconds later.

Downgrading to `BOOTSTRAPPING` rather than `ERROR` is deliberate:
- `running` stays true, so callers **fail closed** through `awaitSelectedRoute` instead of leaking to clearnet,
- the inactivity watchdog re-arms now that `bootstrapPercent < 100`,
- the existing backoff retry gets to re-establish the session, and a successful re-bootstrap clears the tally and restores `RUNNING`.

The thresholds live in `TorCircuitHealthPolicy`, kept free of Android and coroutine dependencies so they are directly testable.

## Evidence

`TorCircuitHealthPolicyTest`, 9 tests, using the log lines from the issue verbatim. Pins both directions — a same-instant burst and occasional failures far apart leave the indicator alone; four failures still going after twenty seconds downgrade it — plus reset and a backwards clock jump.

```
BUILD SUCCESSFUL
```

Measured the same way on `main` and here, the only difference is the new class: **90 classes / 591 tests → 91 / 600**. Lint unchanged from `main` (305 errors, 333 warnings, 17 hints, all pre-existing/baselined).

## What I could not verify

I have no device that reproduces #610, so the wiring inside `ArtiTorManager` — that these Arti lines reach `handleArtiLogLine` and that the indicator visibly turns from connected to connecting — is reasoned from the code and the reporter's log, not observed. The policy itself is covered by tests. The failure markers are matched as substrings of Arti's output, so if Arti changes that wording the detector silently stops firing; that is the main risk in this change and it fails safe (back to today's behaviour).

Happy to adjust the 4-failure / 20-second thresholds if you'd rather have them tighter or looser.

### DIFF
--- a/app/src/main/java/com/bitchat/android/net/ArtiTorManager.kt
+++ b/app/src/main/java/com/bitchat/android/net/ArtiTorManager.kt
@@ -92,6 +92,7 @@ class ArtiTorManager private constructor() {
     private var bindRetryAttempts = 0
     private var inactivityJob: Job? = null
     private var retryJob: Job? = null
+    private var restartJob: Job? = null
     private var currentApplication: Application? = null
     private val circuitHealth = TorCircuitHealthPolicy()
 
@@ -355,6 +356,11 @@ class ArtiTorManager private constructor() {
     }
 
     private fun stopArti() {
+        // Only reached when the user turns Tor off. A pending restart is deliberately not
+        // cancelled by stopArtiInternal — that is what keeps a recovery alive across the stop —
+        // so it has to be dropped here or it would bring Arti back after an explicit off.
+        restartJob?.cancel()
+        restartJob = null
         stopArtiInternal()
         socksAddr = null
         _statusFlow.value = _statusFlow.value.copy(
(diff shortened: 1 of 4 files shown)


Chessing234:
both fair. dropped the "SOCKS connection error" marker — arti-build wraps every handle_socks_connection error in it, including invalid handshake/version/command before a circuit is attempted. the two remaining markers still match the lines in #610.

the restart one is worse than it looks and predates this pr: restartArti stops arti first, and stopArtiInternal cancels retryJob and inactivityJob, so both existing recovery paths cancel themselves between the stop and the start. routing a post-bootstrap outage into that would have traded a wrong icon for a stopped tor. restarts now run on a separate job the stop path leaves alone, cancelled explicitly when tor is switched off.

PR #888, opened 2026-08-16
fix: bound future-dated timestamps on every packet type, not just LEAVE

Chessing234:
## What

`SecurityManager.validatePacket` bounds a packet's timestamp for exactly one type:

```kotlin
if (messageType == MessageType.LEAVE) {
    ...
    if (clockSkew > MESSAGE_TIMEOUT.toULong()) { ... return false }
}
```

ANNOUNCE is bounded separately, in `AnnouncementIdentityValidator` (10 minutes, symmetric). Everything else — MESSAGE, FRAGMENT, FILE_TRANSFER, VOICE_FRAME — is admitted with any timestamp its sender signs. A signature authenticates that the sender *chose* that value; it doesn't sanity-check it.

That timestamp then drives ordering downstream. The one that matters most is `GossipSyncManager.buildGcsPayload`:

```kotlin
list.sortByDescending { it.timestamp.toLong() }
...
val takeN = minOf(nMax, cap, list.size)
```

Broadcasts stamped far in the future sort first and occupy every advertised slot, so real traffic stops being advertised for gossip sync — not just locally, but to every neighbour that syncs against us. The public timeline orders on the same field.

## Fix

Bound the future direction for every packet type, at the allowance that already admits the peer's announcement (`AnnouncementIdentityValidator.MAX_CLOCK_SKEW_MS`, now `internal` so the two cannot drift apart).

Choosing that specific bound matters: a device skewed further than this cannot get an ANNOUNCE verified, so it has no working session today. Nothing that currently works starts failing. LEAVE keeps its tighter symmetric window.

**The past stays deliberately unbounded.** `StoreForward` caches for 12 hours and replays those packets, so a symmetric bound would drop store-and-forward delivery silently.

## Evidence

Four tests added to `SecurityManagerTest` (24 → 28), extending the existing LEAVE replay-window test to the relayed types. Two pin rejection; two pin that the bound is future-only and no tighter than the announcement allowance.

Removing the check fails exactly the two rejection tests:

```
SecurityManagerTest > validatePacket rejects future-dated broadcast messages FAILED
SecurityManagerTest > validatePacket rejects future-dated packets of every relayed type FAILED
28 tests completed, 2 failed
```

The three admission tests — tolerable forward skew, and a six-hour-old message for store-and-forward — pass either way, which is what they are for.

```
BUILD SUCCESSFUL
```

Measured the same way on `main` and here, the only class that changed is `SecurityManagerTest`: **591 → 595 tests**. Lint unchanged from `main`.

## What I could not verify

No on-device or two-radio test — this is verified against `SecurityManager`'s own interface. The judgement call worth a second opinion is the allowance: I tied it to the announcement's 10 minutes rather than the 5 minutes LEAVE uses, on the reasoning above. If you'd rather every type shared LEAVE's 5-minute window, that's a one-constant change, but it would start dropping traffic from peers whose ANNOUNCE we still accept.

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/AnnouncementIdentityValidator.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/AnnouncementIdentityValidator.kt
@@ -8,7 +8,11 @@ import com.bitchat.android.util.toHexString
 
 /** Canonical, side-effect-free preflight for a self-signed mesh announcement. */
 object AnnouncementIdentityValidator {
-    private const val MAX_CLOCK_SKEW_MS = 10 * 60 * 1_000L
+    /**
+     * Clock skew tolerated on a signed announcement. Shared with [SecurityManager] so no packet
+     * type is held to a stricter future bound than the announcement that admits the peer.
+     */
+    internal const val MAX_CLOCK_SKEW_MS = 10 * 60 * 1_000L
 
     fun verify(
         packet: BitchatPacket,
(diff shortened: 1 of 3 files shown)


Chessing234:
awaiting-review

PR #892, opened 2026-08-17
fix(nostr): do not pick relays around (0, 0) for an undecodable geohash

Chessing234:
`RelayDirectory.closestRelaysForGeohash` guards the decode:

```kotlin
val center = try {
    Geohash.decodeToCenter(geohash)
} catch (e: Exception) {
    Log.e(TAG, "Failed to decode geohash")
    return emptyList()
}
```

but `decodeToCenter` never throws. An empty or malformed geohash decodes to the degenerate box at (0, 0), so instead of declining, the app sorts every relay by distance to the Gulf of Guinea and connects to the five closest — `NostrRelayManager.ensureGeohashRelaysConnected` then caches that set for the geohash.

The underlying problem is that the decoder cannot say "this names no cell": (0, 0) is a real place, so `decodeToCenter("")` and `decodeToCenter("s0000")` are indistinguishable to a caller.

- first commit adds `Geohash.isValid`, `decodeToCenterOrNull` and `decodeToBoundsOrNull`, leaving the existing decoders alone (the globe view wants a box back regardless);
- second commit uses the nullable decoder in `RelayDirectory`, so the empty-list path the code already intended is the one that runs.

`./gradlew :app:testDebugUnitTest --tests "com.bitchat.android.geohash.*"` passes locally, including the five new cases in `GeohashValidityTest` (`tests="5" failures="0"`), which cover: base32 accepted in either case, `a/i/l/o` and punctuation rejected, the nullable decoders returning null exactly where the plain ones return the origin, agreement on real cells, and `s0000` — the cell that actually contains (0, 0) — still decoding.

### DIFF
--- a/app/src/main/java/com/bitchat/android/geohash/Geohash.kt
+++ b/app/src/main/java/com/bitchat/android/geohash/Geohash.kt
@@ -67,6 +67,13 @@ object Geohash {
     }
 
     /**
+     * True when every character of [geohash] is a base32 geohash digit and the
+     * string is non-empty, i.e. when it names an actual cell.
+     */
+    fun isValid(geohash: String): Boolean =
+        geohash.isNotEmpty() && geohash.lowercase().all { charToValue.containsKey(it) }
+
+    /**
      * Decodes a geohash string to the center latitude/longitude of its cell.
      * @return Pair(latitude, longitude)
      */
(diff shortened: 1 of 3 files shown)


Chessing234:
on the mesh lab ask — i don't have a second android handset to run the two-device routing session, so i can't produce that result honestly, and i'd rather say so than paste a checklist i didn't run.

what this change actually moves: RelayDirectory.closestRelaysForGeohash previously decoded an invalid geohash to (0, 0) and returned the relays nearest that point in the gulf of guinea. it now returns none for input that doesn't decode, and is unchanged for every geohash that does — decodeToCenterOrNull returns the same center as the old decoder on all valid input, which the added unit tests cover over the full charset. so the routing behaviour for real geohash channels is byte-identical; the only difference is on input that could never have selected the right relays anyway.

happy to hold it for a maintainer with a device pair if you'd still like the mesh lab run on top of that.

Chessing234:
awaiting-review

PR #893, opened 2026-08-17
fix(nostr): accept an uppercase npub, reject a mixed-case one

Chessing234:
BIP-173 says a bech32 string is either all lowercase or all uppercase, and that mixing the two is invalid. `Bech32.decode` looked up each character in the lowercase charset directly:

```kotlin
val index = CHARSET.indexOf(char)
require(index >= 0) { "Invalid character: $char" }
```

so an all-uppercase npub failed on the first character, before the checksum was ever considered. That is the form QR encoders produce — uppercase fits QR's alphanumeric mode, which is denser — so an npub scanned from a QR code or pasted from a site that displays it uppercased could not be decoded.

`ContactIdentityResolver.nostrPubkeyHex` already lowercases at the call site, which is the workaround; `NostrClient.sendPrivateMessage` passes the string through untouched, so that path simply failed.

Now the case is normalised once inside the decoder, after rejecting a mixed-case string outright (which the old code only rejected by accident, since uppercase characters were not in the charset at all).

`./gradlew :app:testDebugUnitTest --tests "com.bitchat.android.nostr.*"` passes locally; the new `Bech32CaseTest` is `tests="4" failures="0"` and covers uppercase decoding to the same payload, mixed case rejected, a corrupted uppercase string still failing on the checksum rather than the charset, and the encode/decode round trip.

### DIFF
--- a/app/src/main/java/com/bitchat/android/nostr/Bech32.kt
+++ b/app/src/main/java/com/bitchat/android/nostr/Bech32.kt
@@ -28,6 +28,13 @@ object Bech32 {
      * alphanumeric mode. Mixed case is invalid and is rejected.
      */
     fun decode(bech32String: String): Pair<String, ByteArray> {
+        // BIP-173 restricts every character to printable US-ASCII, 33..126.
+        // This has to run before the case test and before lowercase(): a
+        // non-ASCII uppercase homoglyph such as U+212A KELVIN SIGN reads as
+        // uppercase, so an otherwise all-uppercase string carrying one shows no
+        // mixed case, and lowercase() then folds it into an ASCII 'k'.
+        require(bech32String.all { it.code in 33..126 }) { "Invalid character" }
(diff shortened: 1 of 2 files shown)


Chessing234:
right, the mixed-case test alone doesn't cover a non-ascii homoglyph. U+212A KELVIN SIGN is uppercase, so an all-uppercase npub carrying one in place of a K passes the check and lowercase() folds it to an ascii k — the string decodes to the same pubkey as the genuine one.

pushed d7e2559: decode now requires every character to be in the printable ascii range 33..126 (what BIP-173 allows) before the case test and before lowercase(). added a test that builds the kelvin-sign variant and expects a rejection; it fails on 29baab6 and passes now. full :app:testDebugUnitTest and :app:lintDebug are green.

Chessing234:
awaiting-review

PR #894, opened 2026-08-17
fix(sync): reject out-of-range GCS parameters instead of decoding garbage

Chessing234:
`GCSFilter.decodeToSortedSet` takes `p` and `m` straight off the wire — a REQUEST_SYNC carries P as a uint8, so anything from 0 to 255 arrives — and decodes with them regardless.

- Kotlin's shift operators use the low 6 bits of the shift count, so `p >= 64` wraps and the decoder emits values that were never in the filter;
- `p = 0` makes every remainder a zero-width read, so the decoder manufactures a value per bit. On a 20-id filter that is **105 fabricated values**.

The decoded set is what `GossipSyncManager.handleRequestSync` consults to decide whether a peer already has a packet, so fabricated values mean the responder **withholds packets the peer actually lacks** — the direction that loses messages.

iOS already guards this, with the reasoning spelled out in `GCSFilter.swift`:

```swift
// Reject out-of-range parameters rather than decoding garbage: callers
// treat the result as "peer has nothing" and fall back to sending data.
guard p >= 1, p <= maxP, m > 1 else { return [] }
```

This is the same guard, with the same bound (`MAX_P = 32`), so both platforms agree on what a filter means.

Verified locally with `./gradlew :app:testDebugUnitTest --tests "com.bitchat.android.sync.*"`:

- before the guard, the new `an out-of-range p decodes to nothing rather than to garbage` fails with `p=0 should decode to nothing, got 105 values`;
- after it, `GCSFilterParameterTest` is `tests="3" failures="0"` and the existing `GCSFilterTest` still passes `tests="3" failures="0"`.

### DIFF
--- a/app/src/main/java/com/bitchat/android/sync/GCSFilter.kt
+++ b/app/src/main/java/com/bitchat/android/sync/GCSFilter.kt
@@ -18,6 +18,14 @@ import kotlin.math.ln
  * - Bitstream is packed MSB-first in each byte.
  */
 object GCSFilter {
+    /**
+     * Highest Golomb-Rice parameter accepted from the wire. P maps to an FPR of
+     * ~1/2^P; beyond 32 the remainder width exceeds any practical filter, and
+     * the shifts in decode silently wrap (Kotlin shifts use the low 6 bits of
+     * the count) into garbage values.
+     */
+    const val MAX_P = 32
+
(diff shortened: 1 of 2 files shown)


Chessing234:
awaiting-review

PR #896, opened 2026-08-17
fix(notes): drop expired notes and match geohash tags case-insensitively

Chessing234:
Two ways `LocationNotesManager.handleEvent` disagrees with the iOS handler on the same event.

**NIP-40 expiry.** iOS drops expired notes client-side, with the reason in the code:

```swift
// NIP-40: relays are not required to enforce expiration — drop
// expired notes client-side so 24h dead drops actually vanish.
let expiresAt = Self.expirationDate(of: event)
if let expiresAt, expiresAt <= self.dependencies.now() { return }
```

Android had no expiration check at all, so an expired note kept showing until the list was cleared or trimmed — the opposite of what a dead drop promises, and the kind of thing a user relies on rather than notices.

**Tag case.** Tag names and geohashes are case-insensitive; iOS lowercases both before matching:

```swift
tag[0].lowercased() == "g" && validGeohashes.contains(tag[1].lowercased())
```

Android compared them exactly (`it[0] == "g"`, `subscribedGeohashes.contains(eventGeohash)`), so a note tagged `["G", "U4PRUYD"]` appeared on iOS and was silently dropped on Android — two devices in the same cell showing different note lists.

Both now follow the iOS handler. The expiration helper mirrors `LocationNotesManager.expirationDate` on that side.

`./gradlew :app:testDebugUnitTest :app:lintDebug` passes locally (rc=0). The handler needs relay and Android dependencies to exercise end to end, so there is no unit test here; the change is a direct transcription of the iOS conditions.

### DIFF
--- a/app/src/main/java/com/bitchat/android/nostr/LocationNotesManager.kt
+++ b/app/src/main/java/com/bitchat/android/nostr/LocationNotesManager.kt
@@ -20,6 +20,19 @@ class LocationNotesManager private constructor() {
         private const val TAG = "LocationNotesManager"
         private const val MAX_NOTES_IN_MEMORY = 500
         
+        /** How often displayed notes are re-checked against their NIP-40 expiry. */
+        private const val EXPIRY_PRUNE_INTERVAL_MS = 60_000L
+
+        /**
+         * Drops notes whose NIP-40 expiry has passed. Their ids stay in
+         * `noteIDs`, so a relay replay cannot resurrect them.
+         */
+        internal fun pruneExpired(notes: List<Note>, nowMillis: Long): List<Note> =
+            notes.filter { note ->
+                val expiresAt = note.expiresAtSeconds ?: return@filter true
+                expiresAt * 1000L > nowMillis
+            }
+
         @Volatile
         private var INSTANCE: LocationNotesManager? = null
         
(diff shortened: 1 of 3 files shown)


Chessing234:
both right, thanks — checked the relay path and the filter gate does run first (NostrRelayManager.handleMessage -> subInfo.filter.matches before dispatch), so the lowercasing in the handler was unreachable over a live subscription.

26100f4: NostrFilter.matches now compares tag names and values case-insensitively. the values this app filters on are hex ids (e, p) or geohashes (g) and both encode the same value in either case, and iOS has no client-side filter at all, so this is the behaviour android was supposed to have. added a test that a filter for u4pruyd accepts ["G", "U4PRUYD"] and still rejects a different geohash — it fails with the old exact comparison and passes now.

15af9bf: notes carry their expiry, and a 60s job started with the subscription and cancelled with it drops the ones that have passed — same interval as the iOS manager's prune timer. ids stay in noteIDs so a replay can't resurrect a dropped note. the pruning itself is a pure function so the boundary case (expiry exactly now) is covered in the tests.

full :app:testDebugUnitTest and :app:lintDebug green.

Chessing234:
awaiting-review

PR #900, opened 2026-08-22
Fix QR scanner crash on camera startup

### LINKED ISSUE #878: App crash when opening QR code scan for verifying user
### Checklist

### Affected version

2.0.1

1. Open the app
2. Click on the top right people icon
3. Press the QR code icon
4. Switch from My QR tab to Scan

### Expected behavior

It should open the Scan tab and show a the camera preview to scan someone else's QR code to verify

### Actual behavior

The app crashes instantly when you try to switch to the scan tab

https://github.com/user-attachments/assets/15c1dc5c-ab62-414e-85cb-1180dce9059d

Android 16

### Affected device model

Nothing phone (2) / A065 

### Additional information

_No response_

phuctoan123:
## Summary
- Guard CameraX provider initialization and retrieval so Scan tab camera startup failures do not escape Compose and crash the app.
- Show the existing QR unavailable message when the camera cannot be started.
- Close the ML Kit scanner when the scanner view is disposed and keep image proxies closed on scanner startup failures.

Fixes #878.

## Testing
- `./gradlew :app:compileDebugKotlin`

Note: full `./gradlew compileDebugKotlin` is blocked locally by a corrupt Android SDK Build Tools 36.0.0 install affecting `:wear:compileDebugKotlin`.

### DIFF
--- a/app/src/main/java/com/bitchat/android/ui/VerificationSheet.kt
+++ b/app/src/main/java/com/bitchat/android/ui/VerificationSheet.kt
@@ -53,7 +53,6 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
-import androidx.compose.ui.draw.clipToBounds
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.asImageBitmap
 import androidx.compose.ui.platform.LocalContext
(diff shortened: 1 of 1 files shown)


Chessing234 (commented):
#899 is changing the same camera-startup path in the same file (VerificationSheet.kt) — it makes the CameraX bindings session-scoped and adds a camera-unavailable recovery, which overlaps what you're guarding here. neither of you will merge cleanly after the other.

yours is the narrower change and the one that matches the crash report, so it may well be the one to take; worth saying so on the thread rather than leaving jack to diff two PRs against each other.

one question on the change itself: closing the ML Kit scanner when the composable leaves is right, but is the close idempotent if the guard already failed during initialization? if the provider future never completed there may be nothing to close, and a throw in the dispose path would be a second crash on the way out.

liuyunming521520-cell:
The overlap callout is correct. I have since split the unrelated location work out of #899 and refreshed #899 onto current `main`. #899 now covers the same startup/disposal failure path with provider/session guards, idempotent analyzer close, and scoped `unbind(preview, analysis)`; it also keeps the embedded viewfinder required by the clipped Crossfade. #900 remains the narrower alternative. To avoid leaving maintainers with two conflicting camera changes, I will not make further overlapping scanner edits; please select one approach for merge rather than applying both.

PR #904, opened 2026-08-23
fix: stop fragmentation eating a signature byte off an unpadded frame

### LINKED ISSUE #903: Fragmentation silently drops about 1 in 256 messages over 512 bytes
### Checklist

### Affected version

2.0.1 (and current main, d55754d)

Reproduced directly against 2.0.1's MessagePadding, driving the same call sequence
FragmentManager.createFragments() uses:

1. Take a frame of 513-768, 1009-1792 or >=2033 bytes ending in 0x01.
2. pad(frame, optimalBlockSize(frame.size))  -> returns the frame unchanged
3. unpad(that)                               -> returns it one byte shorter

Over 200k random frames of 513-4096 bytes, 693 came back truncated (1 in 288 across
that mix; 1 in 256 inside the affected bands). Frames of 769-1008 and 1793-2032 are
unaffected, which is the tell: those are the sizes that actually do get padded.

On air: send >512-byte signed messages repeatedly between two devices and roughly one
in 256 never arrives, on any receiver, with nothing in the sender's log.

### Expected behavior

Every fragmented message reassembles and decodes.

### Actual behavior

createFragments() encodes with packet.toBinaryData() (padded) and then calls
MessagePadding.unpad() to strip the padding before splitting. But across the
fragmentation range no padding was ever applied: pad() returns the data unchanged when
paddingNeeded > 255, and optimalBlockSize() returns the input size unchanged above 2048.

An unpadded frame ends with the last byte of the Ed25519 signature - encode writes the
signature last - and that byte is uniformly random. unpad() reads it as a PKCS#7 pad
length, so when it happens to be 0x01 the validation passes trivially and a real
signature byte is stripped. Every receiver reassembles the fragments correctly and then
fails decodeCore's size check, so the message dies silently on all of them at once.

_No response_

_No response_

### Affected device model

_No response_

### Additional information

One-line fix - ask the encoder for the unpadded frame instead of un-padding afterwards:

```kotlin
val fullData = packet.toBinaryData(padding = false)
```

Byte-identical for frames that actually were padded, so nothing changes on the wire.
iOS doesn't have this: BLEOutboundFragmentPlanner passes the padding flag into
toBinaryData(padding:) and never un-pads.

Hit this while stress-testing long messages in a fork - three receiving devices logged
"Failed to decode reassembled packet" for the same message at the same moment, which is
what pointed at the sender.

Chessing234:
Fixes #903.

`createFragments` encodes the packet with `toBinaryData()` — padding on — and then calls `MessagePadding.unpad()` to strip the padding again before splitting. Across most of the fragmentation range no padding was ever applied:

- `pad()` returns the data unchanged when `paddingNeeded > 255`, and
- `optimalBlockSize()` returns the input size unchanged above 2048.

So frames of 513-768, 1009-1792 and >=2033 bytes come back from the encoder untouched. Such a frame ends with the last byte of the Ed25519 signature, which `encode` writes last and which is uniformly random. `unpad()` reads it as a PKCS#7 pad length, and a value of `0x01` satisfies the tail check trivially — one byte, equal to itself — so a real signature byte is stripped.

Every receiver then reassembles the fragments faithfully and fails the size check in `decodeCore`, which is why the message dies on all of them at once with nothing in the sender's log.

The fix is to ask the encoder for the unpadded frame instead of un-padding afterwards. For frames that really were padded the two are byte-identical, so nothing changes on the wire.

iOS does not have this: `BLEOutboundFragmentPlanner` passes the padding flag into `toBinaryData(padding:)` and never un-pads.

### Evidence

The new test walks the three unpadded bands with a signature ending in `0x01` and compares the reassembled fragment payloads against `toBinaryData(padding = false)`. On `main` it fails exactly as the issue describes:

```
array lengths differed, expected.length=694 actual.length=693;
arrays first differed at element [693]; expected:<1> but was:<end of array>
```

One byte, the trailing `0x01`. The second test pins the padded case so the frame keeps being fragmented without its padding.

`testDebugUnitTest lintDebug` is green. Measured per-class with `--rerun-tasks` on both sides rather than by total, since gradle leaves stale result XMLs behind: `main` is 594 tests in 90 classes, this branch is 596 in 90, and `FragmentManagerTest` is the only class that moved (7 -> 9). No other class changed and there are no failures.

### What I did not run

`AGENTS.md` requires Mesh Lab validation on two physical devices for anything touching fragmentation. I have no devices, so that gate is unmet — someone with the rig should run it before this merges. I also have not reproduced the original symptom over the air; the unit test drives the same call sequence `createFragments` uses, which is the level I can actually verify.

No spec under `docs/` describes the fragment padding step, and the wire format is unchanged, so there is nothing to update there.

This was AI-assisted. The diagnosis in #903 did the hard part, including the one-line fix.

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/FragmentManager.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/FragmentManager.kt
@@ -3,7 +3,6 @@ package com.bitchat.android.mesh
 import android.util.Log
 import com.bitchat.android.protocol.BitchatPacket
 import com.bitchat.android.protocol.MessageType
-import com.bitchat.android.protocol.MessagePadding
 import com.bitchat.android.model.FragmentPayload
 import kotlinx.coroutines.*
 import java.util.concurrent.ConcurrentHashMap
@@ -66,17 +65,16 @@ class FragmentManager {
                 Log.w(TAG, "Rejecting invalid outbound fragment limit: $maxFragments")
                 return emptyList()
             }
-            val encoded = packet.toBinaryData()
-            if (encoded == null) {
+            // Fragment the unpadded frame; each fragment will be encoded (and padded) independently - iOS fix.
+            // Ask the encoder to skip padding rather than un-padding afterwards:
+            // pad() declines a request needing more than 255 bytes and
+            // optimalBlockSize() returns the input size above 2048, so frames of
+            // 513-768, 1009-1792 and >=2033 bytes are never padded. Those end
+            // with the last byte of the signature, and unpad() reads a value of
+            // 0x01 as a one-byte pad and strips a real signature byte.
+            val fullData = packet.toBinaryData(padding = false)
+            if (fullData == null) {
                 Log.e(TAG, "Failed to encode packet to binary data")
-                return emptyList()
-            }
-
-            // Fragment the unpadded frame; each fragment will be encoded (and padded) independently - iOS fix
-            val fullData = try {
-                MessagePadding.unpad(encoded)
-            } catch (e: Exception) {
-                Log.e(TAG, "Failed to unpad data: ${e.message}", e)
                 return emptyList()
             }
 
--- a/app/src/test/java/com/bitchat/android/mesh/FragmentManagerTest.kt
+++ b/app/src/test/java/com/bitchat/android/mesh/FragmentManagerTest.kt
@@ -3,6 +3,7 @@ package com.bitchat.android.mesh
 import com.bitchat.android.protocol.BitchatPacket
 import com.bitchat.android.protocol.MessageType
 import com.bitchat.android.model.FragmentPayload
+import org.junit.Assert.assertArrayEquals
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNotNull
 import org.junit.Assert.assertNull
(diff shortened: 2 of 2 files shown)


NoHomeInSociety:
thanks for picking this up — I have the devices, so here is what I can add on the validation gate.

The symptom first showed up with stock 1.7.4 sending a ~10 KB photo over BLE: 23 fragments, every one delivered, and every receiver failed reassembly. Photos that landed at 10 to 15 fragments went through fine. That was on unmodified upstream fragmentation code, which is what pointed at the padding step in the first place.

I've been running this exact change — `toBinaryData(padding = false)` instead of encode-then-unpad — in a fork since 11 August, on four phones spanning Android 8 to 16. Since then a 45 fragment / 19,872 byte transfer reassembles end to end, a 551 byte message from the fork reassembles on stock 1.7.4 sitting in the same room, and `Failed to decode reassembled packet` hasn't come back in any run.

That's not the Mesh Lab scenario in AGENTS.md and I haven't run it, so the gate is formally still open. But it's physical device evidence on the same code path, with stock builds in the mesh, before and after.

Chessing234:
codex wants mesh lab on devices for the fragmentation frame change — i don't have the hardware here. @NoHomeInSociety if you can re-run the ~10kb / 23-fragment photo case on a build with this branch that would close the gate.

Chessing234:
awaiting-review

Issue #901, opened 2026-08-22
[HIGH] GATT server grants notifications (full mesh broadcast feed) to any unauthenticated peer


diffusionmam:
**Category:** auth-bypass
**Severity:** High
**Primary location:** `app/src/main/java/com/bitchat/android/mesh/BluetoothGattServerManager.kt:285-294` 
**Blast radius:** `BluetoothPacketBroadcaster.kt:428-448`, `BluetoothConnectionTracker.kt:176-179`

---
### Description
The BLE GATT server adds any device that writes the CCCD (`ENABLE_NOTIFICATION_VALUE`) to the
broadcast subscription list: **no authentication, no Noise session, no verified ANNOUNCE, not even
a single prior packet from that device is required**. Once subscribed, the device receives a live
feed of *all* broadcast mesh traffic via `notifyCharacteristicChanged`, plus relayed unicast
traffic (including encrypted DM ciphertext and its routing metadata). An attacker within BLE range
therefore passively harvests identities, social/topology data, public-channel plaintext, and
DM metadata with zero cryptographic work.

This is the transport-layer root of THREAT-MODEL threat **T-01**: *"GATT notifications granted to
any peer; passive receiver of identity + mesh graph."*

---
### code affected
```kotlin
// onDescriptorWriteRequest() — BluetoothGattServerManager.kt:280-299
if (BluetoothGattDescriptor.ENABLE_NOTIFICATION_VALUE.contentEquals(value)) {
    connectionTracker.addSubscribedDevice(device)          // :286  ← unconditional

    connectionScope.launch {
        delay(100)
        if (isActive) {
            delegate?.onDeviceConnected(device)            // :292  ← peer-connect semantics
        }
    }
}
```
Subscription bookkeeping accepts duplicates and has no auth gate:

```kotlin
// BluetoothConnectionTracker.kt:176-179
fun addSubscribedDevice(device: BluetoothDevice) {
    subscribedDevices.add(device)                          // CopyOnWriteArrayList — no contains() check
}
```
Every broadcast/relay path fans out to **all** subscribers:

```kotlin
// broadcastSinglePacketInternal() — BluetoothPacketBroadcaster.kt:428-448
val subscribedDevices = connectionTracker.getSubscribedDevices()
...
subscribedDevices.forEach { device ->
    if (device.address == routed.relayAddress) return@forEach      // only loop-suppression
    if (connectionTracker.addressPeerMap[device.address] == senderID) return@forEach
[...]

PR #906, opened 2026-08-24
fix(ble): make the GATT broadcast subscription idempotent

Chessing234:
### Risk / what this changes

Writing the Client Characteristic Configuration Descriptor is the one GATT operation a peer can repeat freely on an already-open connection: it needs no authentication, no Noise session and no announcement, and the Bluetooth spec places no limit on how often a client may rewrite the descriptor. `onDescriptorWriteRequest` hung two non-idempotent effects off it.

**1. Duplicate entries in the broadcast fan-out list.**

```kotlin
// BluetoothGattServerManager.onDescriptorWriteRequest
if (BluetoothGattDescriptor.ENABLE_NOTIFICATION_VALUE.contentEquals(value)) {
    connectionTracker.addSubscribedDevice(device)      // unconditional append
```
```kotlin
// BluetoothConnectionTracker
fun addSubscribedDevice(device: BluetoothDevice) {
    subscribedDevices.add(device)                      // CopyOnWriteArrayList, no contains()
}
```

`broadcastSinglePacketInternal` iterates that list and calls `notifyDevice` once per entry. N descriptor writes therefore put N entries under one address, and every broadcast the mesh produces afterwards is transmitted to that address N times — N unbounded, chosen by the peer, paid for in this device's radio time and battery.

**2. An unthrottled mesh-wide ANNOUNCE per write.**

The same branch scheduled `delegate?.onDeviceConnected(device)`, and `BluetoothMeshService.onDeviceConnected` calls `sendBroadcastAnnounce()`. That has no throttle and floods at `maxTtl`, so a peer could drive a mesh-wide ANNOUNCE at whatever rate it rewrote the descriptor — the amplification reaches past the attacker's own link into the whole mesh.

**3. `DISABLE_NOTIFICATION_VALUE` was never handled.** A client that unsubscribes per spec stayed on the broadcast feed until it disconnected.

### The fix

- `addSubscribedDevice` is idempotent by address and returns whether the device was newly subscribed. Peer-connect side effects run only on that transition.
- `DISABLE_NOTIFICATION_VALUE` unsubscribes instead of being ignored.
- `removeSubscribedDevice` matches on address like the connection-cleanup paths already do, so an unsubscribe cannot miss because the stack handed us a different `BluetoothDevice` instance for the same address.
- The descriptor-value classification moves into `GattSubscriptionPolicy` so the decision itself is unit-testable.

`docs/device_manager.md` §3 (GATT Server) gains the corresponding rule.

### Scope — what this deliberately does not do

This is the bookkeeping half of #901. It does **not** add the authentication gate that issue proposes for granting notifications.

That gate is a protocol design call for the maintainer, and I did not want to make it inside a bug fix: a client normally subscribes right after service discovery, before it has written anything, so failing closed on "no verified announce yet" risks breaking mesh bootstrapping outright — and an attacker can satisfy an announce gate anyway, since the first announce is self-signed TOFU by design (`docs/NOISE_PEER_ID_BINDING.md`, "Remaining TOFU boundary"). The amplification fixed here is real and unconditional regardless of how that question is settled.

### Tests

`./gradlew testDebugUnitTest lintDebug` (the CI task pair) is green on this branch, which is cut directly from `main`, so head and the merge commit are the same tree.

New `GattBroadcastSubscriptionTest`: **9 tests, 0 failures**. It has teeth — restoring the previous `add`/`remove` bodies fails 3 of them:

```
GattBroadcastSubscriptionTest > repeated enables subscribe an address exactly once FAILED
    java.lang.AssertionError: a repeat enable is not a new subscription
GattBroadcastSubscriptionTest > a second framework instance for the same address does not add a second entry FAILED
GattBroadcastSubscriptionTest > unsubscribing removes the address whichever instance carries it FAILED
```

Lint: the 14 findings on the two touched files are all pre-existing `MissingPermission` on lines outside this diff; the new file adds none.

**Not run: Mesh Lab.** This touches transports, so the physical-device gate in `AGENTS.md` applies and I have no two-device rig — the on-air behaviour of a repeat CCCD write is verified by reading the fan-out path, not by measurement. Worth confirming there before merge.

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/BluetoothConnectionTracker.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/BluetoothConnectionTracker.kt
@@ -171,18 +171,39 @@ class BluetoothConnectionTracker(
     }
     
     /**
-     * Add a subscribed device
+     * Subscribe a device to broadcast notifications.
+     *
+     * Idempotent by address: a peer may rewrite the notification descriptor as
+     * often as it likes, and each extra entry here would notify that address one
+     * more time for every broadcast the mesh produces. Returns true only when
+     * the device was not already subscribed, so callers can keep peer-connect
+     * side effects tied to an actual new subscription.
      */
-    fun addSubscribedDevice(device: BluetoothDevice) {
-        subscribedDevices.add(device)
-    }
+    fun addSubscribedDevice(device: BluetoothDevice): Boolean =
+        synchronized(connectionStateLock) {
+            if (subscribedDevices.any { it.address == device.address }) {
+                false
+            } else {
+                subscribedDevices.add(device)
+                true
+            }
+        }
     
     /**
-     * Remove a subscribed device
+     * Remove a subscribed device. Matches on address, like the connection
+     * cleanup paths, so an entry cannot outlive an unsubscribe because the
+     * framework handed us a different BluetoothDevice instance for the same
+     * address.
      */
     fun removeSubscribedDevice(device: BluetoothDevice) {
-        subscribedDevices.remove(device)
+        synchronized(connectionStateLock) {
+            subscribedDevices.removeAll { it.address == device.address }
+        }
     }
+
+    /** Whether this address currently receives broadcast notifications. */
+    fun isSubscribed(deviceAddress: String): Boolean =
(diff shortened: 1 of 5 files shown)


Chessing234:
awaiting-review

PR #908, opened 2026-08-25
fix(ble): require a verified ANNOUNCE before granting broadcast notifications

Chessing234:
## Summary
- Reject CCCD enable requests with `GATT_INSUFFICIENT_AUTHORIZATION` until the central has sent a signature-valid ANNOUNCE on that link
- Record eligibility when `MessageHandler` accepts an announce and wire it through `BluetoothConnectionManager`
- Unit-test the eligibility policy

## Risk / scope
This is the participation gate half of #901. It deliberately does not replace #906 (idempotent subscription bookkeeping); the two stack cleanly.

Boot order: a normal peer discovers services and may write the CCCD before its first announce lands. Those writes fail closed until the announce is verified, then the client can retry the descriptor write per the BLE spec.

## Test plan
- [ ] `./gradlew testDebugUnitTest lintDebug` (could not run locally — JDK 21 toolchain missing here; CI covers it)
- [ ] Mesh Lab: confirm a real peer still completes discovery after its first announce

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/BluetoothConnectionManager.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/BluetoothConnectionManager.kt
@@ -101,6 +101,10 @@ class BluetoothConnectionManager(
     // Public property for address-peer mapping
     val addressPeerMap get() = connectionTracker.addressPeerMap
 
+    fun noteVerifiedAnnounceReceived(deviceAddress: String) {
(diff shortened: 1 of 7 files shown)


Chessing234:
deferred the cccd instead of rejecting it — android writes that during discovery and never retries, so a gatt reject left them unsubscribed even after the announce. the feed still stays withheld until a verified announce lands.

Chessing234:
awaiting-review

PR #910, opened 2026-08-25
fix: reject a future-dated verification QR

Chessing234:
## Summary
- `verifyScannedQR` treated freshness as `now - ts`. A timestamp in the future made that negative, so a QR stayed valid longer than the 5-minute window.
- iOS already uses `abs(now - ts)`. Use the same skew check on Android.

## Test plan
- [x] `./gradlew :app:testDebugUnitTest --tests com.bitchat.android.services.VerificationServiceTest` locally (fresh, future-dated, and expired signed payloads)
- [ ] remaining unit tests / lint on CI


Chessing234:
awaiting-review

PR #911, opened 2026-08-25
fix: refuse an ambiguous nickname instead of picking the first peer

Chessing234:
## Summary
- `/msg`, `/block`, and geohash DMs resolved a nickname with the first map entry. Two peers named alice meant the command could hit the wrong person (Cure53 BCH-01-003 / #589).
- Mentions already omit an ambiguous base. Use the same unique-or-nil rule, and require the people-list `#xxxx` suffix when names collide.

## Test plan
- [x] `./gradlew :app:testDebugUnitTest --tests com.bitchat.android.ui.ChatUIUtilsTest` locally (unique, colliding, mesh first-four suffix, geohash last-four suffix)
- [ ] remaining unit tests / lint on CI

### DIFF
--- a/app/src/main/java/com/bitchat/android/nostr/GeohashRepository.kt
+++ b/app/src/main/java/com/bitchat/android/nostr/GeohashRepository.kt
@@ -40,10 +40,12 @@ class GeohashRepository(
 
     @Synchronized
     fun findPubkeyByNickname(targetNickname: String): String? {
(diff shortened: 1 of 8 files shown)


Chessing234:
people list and @-mentions now show the same #xxxx suffix the resolver accepts, so /msg alice#1111 is actually on the row.

Chessing234:
awaiting-review

PR #914, opened 2026-08-26
Select geohash relays the way iOS selects them

heyaim:
Both platforms pick the five relays nearest a geohash and use only those. A geohash message crosses platforms only if the sender's five and the receiver's five share a relay. Android fetched `nostr_relays.csv` from the georelays repo; iOS fetches `relays/online_relays_gps.csv` from the bitchat repo, a copy that changes only when a maintainer merges the automated update, and it has sat at its Jul 26 snapshot since (permissionlesstech/bitchat#1617). Measured Aug 11 over 6,013 points against iOS's own selection code, the two selections shared a mean of 2.3 of 5 relays, and about one point in nine shared none. A test event published to the old Android five reached iOS through exactly one relay, the shared one.

## Changes

- **Fetch the directory from the file iOS reads**, and point the weekly `fetch-georelays.yml` update at the same file. The bundled asset is not touched here; the job refreshes it from the new source, and a fresh install uses the bundled list only until its first download completes.
- **Key each row by the host string iOS builds**: lowercased, an explicit port kept unless it is 443. 115 hosts appear twice in the directory, bare and with `:443` (230 of 441 rows), and both copies could fill nearest-5 slots. This is a bug on main independent of which file is read.
- **Order distance ties by that key**, as `GeoRelayDirectory.closestRelays` does; 385 of 441 rows share a coordinate with another, and the file's row order decided ties before.

With all three, both platforms select the same five relays in the same order at all 6,013 points. One difference remains: iOS rejects a whole file on a malformed row while this app skips the row; #917 addresses that.

## Tests

Eight cases in `RelayDirectoryTest`: the fetch URL, the host key with and without `:443`, dedup, tie ordering from a deliberately shuffled file, and five distinct servers for a directory whose nearest relay is listed twice. Full suite, lint and build pass. On an emulator the app loads the asset, then fetches and swaps to the download.

## Notes

- Both platforms now share the bitchat copy's staleness instead of splitting on it; aligned and stale still delivers where fresh and split does not. The staleness itself is fixed upstream, by that repo's update landing.
- On a duplicated host with conflicting coordinates this app keeps the first row while iOS rejects the file; no current host has one.
- An install holding the old download converges within 24 hours.
- The directory remains an unsigned third-party trust anchor (docs/security-review-jul-27.md M9); the fetch moves to the bitchat repo's copy, which its maintainers review before it lands.

### DIFF
--- a/.github/workflows/fetch-georelays.yml
+++ b/.github/workflows/fetch-georelays.yml
@@ -19,9 +19,11 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
       
       - name: Fetch GeoRelays
+        # Must match RelayDirectory.ASSET_FILE_URL, the file the app fetches at
+        # runtime; the bundled asset is a snapshot of the same file.
         run: |
-          wget https://raw.githubusercontent.com/permissionlesstech/georelays/refs/heads/main/nostr_relays.csv
-          mv nostr_relays.csv ./app/src/main/assets/nostr_relays.csv
+          wget https://raw.githubusercontent.com/permissionlesstech/bitchat/refs/heads/main/relays/online_relays_gps.csv
+          mv online_relays_gps.csv ./app/src/main/assets/nostr_relays.csv
 
       - name: Check for changes
         id: git-check
--- a/app/src/main/java/com/bitchat/android/nostr/RelayDirectory.kt
+++ b/app/src/main/java/com/bitchat/android/nostr/RelayDirectory.kt
@@ -26,7 +26,13 @@ import okhttp3.Request
 object RelayDirectory {
 
     private const val TAG = "RelayDirectory"
-    private const val ASSET_FILE_URL = "https://raw.githubusercontent.com/permissionlesstech/georelays/refs/heads/main/nostr_relays.csv"
+
+    // The same file iOS reads (GeoRelayDirectory.swift). Both platforms take the 5
+    // nearest relays from their directory and use them without the defaults, so a
+    // geohash message crosses platforms only if the two selections share a relay.
+    // Reading different files made the selections diverge. Selecting from the same
+    // file, with rows keyed and ordered the same way, keeps them aligned.
+    internal const val ASSET_FILE_URL = "https://raw.githubusercontent.com/permissionlesstech/bitchat/refs/heads/main/relays/online_relays_gps.csv"
     private const val ASSET_FILE = "nostr_relays.csv"
     private const val DOWNLOADED_FILE = "nostr_relays_latest.csv"
     private const val PREFS_NAME = "relay_directory_prefs"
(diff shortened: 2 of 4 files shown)


Chessing234 (commented):
i checked the parity claim against the iOS implementation rather than the description, and the good news first: on the file as committed here, the two agree exactly. i ran iOS's `validatedDirectoryAddress` rules over all 441 rows of `nostr_relays.csv` and compared to this branch's `canonicalHost` — zero rejected rows, zero canonicalisation mismatches, zero hosts carrying conflicting coordinates. the tie-break reasoning is right too: iOS sorts by `(distance, entry.host)` and emits `"wss://\(host)"`, and since every url here gets the same `wss://` prefix, ordering by url reproduces it exactly.

what isn't aligned is the **rules**, only today's data. `GeoRelayDirectory.parseCSV` is far stricter than `parseCsv` here, and the difference is invisible until a row changes:

- **iOS rejects the whole file on one bad row.** every `guard` in its parse loop is `return nil`, and the caller keeps the previous directory. android `continue`s past a bad row and accepts the rest. so a single malformed line published to the shared csv leaves iOS on yesterday's 441 relays and android on today's 440 — a bigger divergence than the two-file situation this pr is fixing, because it hits every geohash at once and neither side surfaces it.
- **iOS validates the header exactly** against `["relay url","latitude","longitude"]` or `["relay url","lat","lon"]` and rejects anything else. android just skips any line starting with `relay url` and parses on.
- **`validatedDirectoryAddress` rejects things `canonicalHost` accepts**: non-ASCII, userinfo, query, fragment, a path other than `/`, `localhost` / `.local` / `.internal`, single-label hosts, all-numeric hosts, labels outside 1–63 chars or outside `[a-z0-9-]`, hosts over 253 chars, ports outside 1–65535. `canonicalHost` splits on `://` and the last `:` and lowercases. any row that trips one of those is dropped by iOS and kept by android, which shifts the nearest-5 for every geohash near it.
- **duplicate host with different coordinates**: iOS rejects the file ("one endpoint cannot truthfully occupy two coordinates"); android's `seenEndpoints` keeps whichever row came first.

given the pr's own argument — that delivery depends on both clients selecting from *identical data* — the filter is the third input alongside the file and the ordering, and it's the one still unmatched. porting the reject-the-file semantics and the host validation would close it, and it's mostly mechanical.

two smaller things:

the bare-vs-`:443` dedup is a real bug and worth calling out separately from the parity work, because it's a bug on main today independent of which file is read. but note iOS gets there differently: `validatedDirectoryAddress` strips the `:443` so both rows produce the same host key, and `entriesByHost` then *rejects the file* if their coordinates differ rather than keeping the first. worth matching that too, since the duplicate rows are exactly the ones most likely to have drifted coordinates.

and `ASSET_FILE` is still `nostr_relays.csv` while the remote is now `online_relays_gps.csv`. that works, but the bundled asset and the cache file now carry a name that no longer describes where they come from, and iOS's `bundledCSVURLs` looks for both names for the same reason. worth renaming or leaving a comment so the next person doesn't go looking for a georelays fetch that isn't there any more.

heyaim:
This PR makes both platforms read the same file and select from it the same way; the acceptance rules are the follow-up it was fenced around, and that follow-up is built: [#917](https://github.com/permissionlesstech/bitchat-android/issues/917), stacked on this branch. It ports GeoRelayDirectory's whole-file validation rule for rule, adds cache invalidation for upgraded installs, and is cross-checked against the real implementation; the check surfaced five places iOS behaves differently from its own source text, all handled and pinned. It is a draft until this merges, so each PR stays small and easy to review.

Thanks for the independent check on the parity.

The asset and cache names predate the URL move; a comment now records that both are copies of online_relays_gps.csv.

Chessing234 (commented):
selection alignment looks real, but the workflow still hand-refreshes `app/src/main/assets/nostr_relays.csv` in the same PR as the selector change. if the fetch URL is the source of truth, keep the bundled snapshot update on the automation path (or a follow-up that only bumps the asset from a verified sha) so a selector bug and a directory bump aren't one review. also call out the remaining parse-strictness skew (iOS rejects the whole file, android skips bad rows) in the PR title/body as an explicit non-goal — easy to miss under the "same five" claim.

heyaim:
Rebased onto main after the Sep 6 relay-data commit conflicted with the bundled asset; the asset is now out of this PR. Also corrected two numbers in the description to match the recorded 6,013-point run: 2.3 of 5 relays shared on average, and about one point in nine sharing none.

heyaim:
@Chessing234 Both done. The bundled asset is out of this PR; the workflow commit points the weekly job at the file the app fetches, and the snapshot now moves on the automation path by itself. The parse-strictness difference is stated in the body as a non-goal, with #917 as the follow-up.

PR #915, opened 2026-08-26
Refactor UI root out of MainActivity

yet300:
# Description
This PR extracts the Compose app root from MainActivity into a separate BitchatApp composable.

The goal is to reduce the responsibilities of MainActivity and make it closer to a lightweight Android entry point. Right now it handles Android lifecycle/service setup together with the main Compose UI tree, onboarding rendering, and chat rendering.

This is intentionally a small, behavior-preserving refactor. It does not introduce Navigation 3 or change the existing onboarding, mesh, ViewModel, or back-navigation behavior.

This also prepares the UI structure for a gradual Navigation 3 migration in follow-up PRs without requiring a large architectural rewrite at once.


Chessing234 (commented):
this extraction restores `addCallback` from the composable path. its `DisposableEffect` is keyed only by the activity owners, so it survives recompositions, but it still installs the callback while the chat branch is active; #912 replaces this exact pattern because the old callback accumulated on each recomposition. please carry that `BackHandler` fix into `BitchatApp` rather than reintroducing it here.

Chessing234:
correction: `DisposableEffect` keys this callback to the lifecycle owners and removes it when the chat branch leaves composition, so it does not accumulate across recompositions. my review above was incorrect; please disregard it.

Chessing234 (commented):
beyond the #912 overlap: moving setContent into BitchatApp means any Activity-scoped things that previously closed over MainActivity (permission launchers, onNewIntent wiring, edge-to-edge) need a hard audit — grep for `this@MainActivity` / LocalContext casts that assumed the old structure. also the PR is large enough that a smoke checklist (cold start, onboarding → chat, back from dm, process death) should be in the body before merge.

PR #917, opened 2026-08-26
Validate relay directories with iOS's rules

heyaim:
Both clients pick their geohash relays from the same CSV directory, and each parses it for itself. iOS applies strict rules to that file: one malformed or conflicting row rejects the whole directory and the previous copy stays, the header must match exactly, size, rows and entries are capped, and every host is screened. This app skipped bad rows and accepted almost any host. This PR ports iOS's rules; both platforms then accept the same files, as #914 makes them read the same file and select the same way. The bitchat repo already validates the file before it lands, making a bad row unlikely; the port bounds what the client accepts regardless. Stacked on #914 until it merges.

## Changes

- **`parseCsv` and `canonicalHost` are replaced by a port of `GeoRelayDirectory.validatedEntries` and `validatedDirectoryAddress`**, with the live limits: 512KB, 5,000 rows and entries, a 50-entry floor for downloads. One malformed or conflicting row rejects the file; the header check, coordinate ranges and host screen match iOS.
- **A rejected file keeps the previous directory.** A corrupted cache falls back to the bundled asset, which passes the same gate.
- **The baseline-overlap guard is ported**: a download that keeps less than half of the known entries is rejected even when well formed.
- **Five details match iOS's compiled validator where it differs from a plain reading of the Swift source**, each with a test: a single BOM accepted, percent escapes decoded after parsing, hex coordinates, negative zero equal to zero on duplicate rows, and punycode labels refused. iOS reaches the punycode verdict by decoding; this port refuses the label outright.
- **The download cache remembers its source URL**, and an install upgraded across #914's URL change refetches at first launch instead of at the next daily check.
- **The acceptance rules join the selection rules in `docs/client-rewrite-contracts.md`.**

## Tests

Thirty-four cases in the relay suite, 26 new: header forms, whole-file rejection, the limits, the five runtime details, the host screen in both directions, the overlap guard both ways, the bundled asset passing validation, and a committed snapshot yielding exactly 326 entries; the count lives on the snapshot because the weekly update rewrites the asset. Full suite, lint and build pass. Cross-checked against the Swift validator on 84 fixture runs and 100,000 fuzzed hosts and coordinates: 15 of 50,000 hosts differ, all refused here and salvaged by Foundation, 13 of them punycode. On an emulator: a fresh install loads the asset through the validator, a corrupted cache falls back to it, a cache from the old source is dropped at startup, and a download validates and swaps in.

## Notes

- A draft until #914 merges.
- A rejected download leaves the update stamp unset, and the existing one-minute staleness check refetches until the file changes; iOS backs off. Not changed here.
- Duplicate rows with conflicting coordinates now reject the file, as on iOS; #914 kept the first row. No current host has one.
- Transport and connection behavior are untouched.


Chessing234 (commented):
looked at the validator port and the cache invalidation on source_url change — both look aligned with the iOS rules and the tests pin the rejection cases. one thing to double-check before merge: the bundled `nostr_relays.csv` rewrite is a large hand-edited snapshot of the new source — worth confirming it matches a fetch of `online_relays_gps.csv` at the same tip so the first-run asset and the remote file don't diverge on day one.

heyaim:
The bundled asset is a fetched copy, not a hand edit: 1006e48 snapshotted online_relays_gps.csv byte for byte, and a fetch of the same URL on Aug 30 matched it exactly, sha256 811523100064820b024714eaf3bbe0a9ba99a3c257e453e01bc1bc0f3bf8401a.

The check turned up one divergence risk outside the file: fetch-georelays.yml downloads the relay list from the georelays repo every week and pushes it straight to main. Two of the new commits close it: the job now downloads the same file the app reads, and the exact-count test here moves to a committed snapshot.

Thanks for the check.

Issue #919, opened 2026-08-28
[Feedback + Bug Report] BitChat Android 2.0.1 — globe geohash view, geochat participant counter, note persistence, Tor default routing, channel/Mesh isolation, relay sync logging


nomadram:
# 0. Preface

Before filing this, I searched open and closed issues for these topics and didn't find existing reports covering the same combination. Apologies if any of this overlaps with something already tracked, feel free to close/merge if so.
Thanks for the work in this release. The UI/UX is a lot better than earlier versions.
Thanks in particular for the reworked geohash map, now a native animated 3D globe. Good addition, easier to use than the old flat view.

# **1. Environment**

App: BitChat Android 2.0.1
Device: OnePlus Nord 5
OS: Android 16, OxygenOS 16

# 2. Issues

**2.1 Geochat participant counter is unreliable**
Symptom:
Displayed participant count for a geochat is inconsistent. Often shows "1" or no count at all, regardless of actual chat activity.
The counter also changes on its own, without any visible connection/disconnection event tied to it.
Assumption, not confirmed: could be a stale/cached subscription count from the relay, or a race condition between the UI layer and the peer-list update. No logcat to back this up yet, can provide if useful.

**2.2 Notes on a geohash are not persisted**
Symptom:
Notes attached to a specific geohash don't save. Not visible on the next visit to that geohash, at least on the originating device.
Assumption, not confirmed: persisting a note might require acknowledgment/replication from more than one relay or device (not just a local write, but confirmation the note propagated beyond the originating device) before it counts as saved. If that's the intended design, it isn't communicated to the user anywhere right now, no in-app message, no docs, so from the user side it just looks broken. If this assumption is wrong and it's a plain persistence bug, disregard the caveat.

**2.3 Suggestion: Tor routing should default to off, with support for user-configured bridges**
Context: in countries where Tor is blocked at the network/ISP level, Tor routing enabled by default can block the app from connecting at all, with no indication to the user why.
Suggestion:
Ship with Tor routing disabled by default (opt-in, not opt-out).
Add a settings option for user-configured Tor bridges, so users who need Tor in a restrictive network can supply their own bridge lines and get it working, instead of being stuck with a default relay set that doesn't connect.
This is a feature request more than a bug, including it here since it's tied to real-world usability of Tor routing as currently shipped.

**2.4 Channels still leak content into the Mesh chat category**
Symptom:
[...]

PR #920, opened 2026-08-29
Carry channel messages in the BitchatMessage binary envelope

Chessing234:
## Summary
- `MeshCore` / `BluetoothMeshService.sendMessage` ignored `channel` and put raw UTF-8 on the wire, so every `#room` message also appeared in Mesh (#919 §2.4).
- Channel sends now use `BitchatMessage.toBinaryPayload()`; public mesh stays UTF-8 for iOS interop.
- `MessageHandler` prefers `fromBinaryPayload` so `IncomingMessageAdmission` can route by `channel`.

## Test plan
- [ ] `ChannelWireFormatTest` round-trip
- [ ] Send in a joined channel on two Android devices — message stays in the channel, not Mesh
- [ ] Public mesh message still arrives as plain text (including from iOS)

Refs #919

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/BluetoothMeshService.kt
@@ -852,6 +852,34 @@ class BluetoothMeshService(private val context: Context) : TransportBridgeServic
         return reusable
     }
     
+
+    /**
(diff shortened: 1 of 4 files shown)


Chessing234:
moving this to draft: the current receive path still preserves envelope-controlled sender/private fields, and failed channel encoding still falls back to public UTF-8. both review findings need transport-level regression coverage before this is ready for review.

Chessing234:
awaiting-review

PR #921, opened 2026-08-29
Raise NIP-17 gift-wrap fetch limit from 100 to 1000

Chessing234:
## Summary
- `NostrFilter.giftWrapsFor` used `limit = 100`. Relays return at most that many stored events, newest by randomized `created_at`.
- One delivered+read DM is typically three gift wraps, so 100 ≈ 33 messages inside the lookback — and loss is permanent across restarts.
- Raise to 1000 (same budget as geohash message fetches).

## Test plan
- [ ] Cold-start with a busy DM inbox; more historical gift wraps arrive than before
- [ ] Live gift wraps after EOSE still flow

Refs #858


Chessing234:
awaiting-review

PR #922, opened 2026-08-29
Fall back when encrypted channel send is unimplemented

Chessing234:
## Summary
- `ChannelManager.sendEncryptedChannelMessage` returned without calling `onEncryptedPayload` or `onFallback`.
- Callers only transmit from those callbacks, so password-protected channel sends never left the device.
- Invoke `onFallback()` until the AES path is restored end-to-end.

## Test plan
- [ ] Send in a password-protected channel — message reaches peers via the fallback path
- [ ] Local echo still appears in the channel timeline

Refs #919


Chessing234:
closing this because the fallback would broadcast password-protected channel content as plaintext. preserving fail-closed behavior is safer until encryption is restored.

PR #925, opened 2026-09-03
Mark REQUEST_SYNC responses with the RSR flag

heyaim:
This does the protocol part of #622. When a peer answers a REQUEST_SYNC it replays packets from its archive with their original timestamps, and as that issue says, the only thing marking a reply is TTL=0, so a receiver applying a freshness window filters the replay out.

iOS added flag 0x10 for this in permissionlesstech/bitchat#965 (January), merged as the symmetric half of Android #628, which never landed; permissionlesstech/bitchat#998 then removed iOS's legacy acceptance of unflagged TTL=0 replies. Since then iOS drops every Android sync reply older than 120 seconds, while it accepts flagged replies up to six hours old. An iOS user joining a room of Android phones gets two minutes of history instead of what those phones' archives hold.

## Changes

- **Define flag `0x10` as `IS_RSR`** in `BinaryProtocol.Flags`, matching the bit iOS already uses, and carry it through encode and decode as `BitchatPacket.isRSR`. It stays out of `equals`/`hashCode`, as `PacketIdUtil` computes identity from type, sender, timestamp and payload alone.
- **Set it at the two archive replay sites** in `GossipSyncManager`, alongside the TTL change already made there, and **carry it into fragments**: `FragmentManager` builds each fragment with the original packet's timestamp, so an unmarked fragment of a replay would be judged on that old timestamp and dropped. iOS propagates the flag the same way.
- **Keep it out of the signing preimage**, as TTL already is: a packet is signed at first send and marked when later served, so including it would invalidate every replayed signature. iOS's preimage hardcodes it false the same way.
- **The REQUEST_SYNC spec (`docs/sync.md`) now names the flag.**

This is send-side only. Nothing changes about what this app accepts.

## Tests and validation

Twelve cases: `BinaryProtocolTest` pins the wire bit, the untouched signing preimage, an end-to-end Ed25519 sign, mark and verify, the byte-reuse interaction with #863, and that unknown flag bits stay ignored on decode; `GossipSyncRsrFlagTest` pins both replay sites, each separately, and that archiving does not mark; `FragmentManagerTest` pins fragment propagation. `./gradlew testDebugUnitTest lintDebug assembleDebug`: 820 total, 817 passed, 3 skipped (the pre-existing ignores, also skipped on main), 0 failures. Not validated on hardware.

## Limits

This change stands on its own. iOS has accepted flagged replies that fit in one frame since January. Single-frame messages, which covers ordinary chat text, are fully restored as soon as an Android peer sets the flag.

A reply too long for one frame is only partly recovered: on reassembly iOS validates the inner packet against its original author, not the serving peer, so it lands when the author is the responding peer or one iOS has an open sync request with, and is dropped otherwise. permissionlesstech/bitchat#1705 changes the iOS side to judge the reassembled packet by the peer that served it. Until that ships, a fragmented replay of a recent third-party message is validated against its author instead of passing on its timestamp.

One more case matches iOS-to-iOS behavior instead of slipping through on a fresh timestamp: a reply to iOS's discovery-phase broadcast request waits for the next round, since iOS registers only requests sent to a specific peer, and the same history is served and accepted then.

## Notes

- Draft PR #720 edits the same two lines in `handleRequestSync`, and #885 wraps the same function in a per-peer budget; whichever lands later rebases, and `GossipSyncRsrFlagTest` fails if the flag is dropped in any resolution. The flag was built before with a receive side in #628 and #650, both closed unmerged; this PR takes only the send half, which iOS's merged receive side already accepts.

### DIFF
--- a/app/src/main/java/com/bitchat/android/mesh/FragmentManager.kt
+++ b/app/src/main/java/com/bitchat/android/mesh/FragmentManager.kt
@@ -150,7 +150,11 @@ class FragmentManager {
                     timestamp = packet.timestamp,
                     payload = fragmentPayload.encode(),
                     route = packet.route,
-                    signature = null // iOS: signature: nil
+                    signature = null, // iOS: signature: nil
+                    // Fragments carry the original packet's timestamp, so a receiver applying a
+                    // freshness window judges each fragment on it. A replayed archive packet
+                    // large enough to fragment would be dropped without this.
+                    isRSR = packet.isRSR
                 )
 
                 fragments.add(fragmentPacket)
--- a/app/src/main/java/com/bitchat/android/protocol/BinaryProtocol.kt
+++ b/app/src/main/java/com/bitchat/android/protocol/BinaryProtocol.kt
@@ -64,7 +64,8 @@ class WirePayload(
  * - Type: 1 byte
  * - TTL: 1 byte
  * - Timestamp: 8 bytes (UInt64, big-endian)
- * - Flags: 1 byte (bit 0: hasRecipient, bit 1: hasSignature, bit 2: isCompressed)
+ * - Flags: 1 byte (bit 0: hasRecipient, bit 1: hasSignature, bit 2: isCompressed,
+ *   bit 3: hasRoute (v2+), bit 4: isRSR)
  * - PayloadLength: 2 bytes (v1) / 4 bytes (v2) (big-endian)
  *
  * Variable sections:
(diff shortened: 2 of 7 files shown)


Chessing234 (commented):
send-side marking and the wire bit look right, and keeping isRSR out of the signing preimage matches iOS. what's missing is the receive path: SecurityManager already applies a freshness window, and this PR's docs say RSR exempts that, but nothing here checks that we actually asked the delivering peer (iOS's RequestSyncManager.isValidResponse). without that gate, anyone can set 0x10 and slip stale signed traffic past the skew check once exemption lands — or iOS-origin RSR replies keep getting dropped on android today. wire the solicitation check (and a reject for unsolicited isRSR) in the same PR as the flag.

heyaim:
The receive path you describe is iOS's, not Android's: the skew check in SecurityManager's validatePacket covers LEAVE only, a public MESSAGE has no age bound, and no check on the receive path reads the flag, which means nothing is bypassed by setting it and no iOS reply is dropped for carrying it. iOS has had that check since January (permissionlesstech/bitchat#965, permissionlesstech/bitchat#998): BLEIngressPacketGuard rejects a flagged packet with no open request to the delivering peer. The exemption and that check belong together, on the receive side.

PR #927, opened 2026-09-04
Reject implausible Nostr DM rumor and gift-wrap timestamps

Chessing234:
## Summary
- Client-side window on the **inner rumor** `created_at`: reject timestamps outside `[now − 24h − 15min, now + 15min]`, matching iOS `NostrInboundPipeline.isPlausibleRumorTimestamp`.
- Reject **future-dated outer gift wraps** the same way (previously a negative age passed the one-sided `messageAge > 173700` check).
- Outer gift-wrap age ceiling stays 48h + 15min so existing NIP-17 randomization still delivers.

A relay that ignores the subscription `since` filter can otherwise inject arbitrarily old or future-dated DMs; gift-wrap age alone is not enough because NIP-17 randomizes only the outer envelope.

## Test plan
- [x] Unit tests for rumor and gift-wrap windows (`NostrTimestampPolicyTest`)
- [ ] CI `testDebugUnitTest`
- [ ] Could not run full suite locally (no Android SDK in this environment); CI covers it

### DIFF
--- a/app/src/main/java/com/bitchat/android/nostr/NostrClient.kt
+++ b/app/src/main/java/com/bitchat/android/nostr/NostrClient.kt
@@ -241,10 +241,8 @@ class NostrClient private constructor(private val context: Context) {
         giftWrap: NostrEvent,
         handler: (content: String, senderNpub: String, timestamp: Int) -> Unit
     ) {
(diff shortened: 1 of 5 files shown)


Chessing234:
awaiting-review

PR #928, opened 2026-09-04
Reject odd-length and 0x-prefixed hex the way iOS does

Chessing234:
## Summary
- Match iOS hex parsing: trim whitespace, remove an optional `0x`/`0X` prefix, reject odd lengths, and preserve empty input as empty bytes.
- Reject signed and non-ASCII chunks before radix conversion so inputs such as `0x-1`, `+1`, and full-width digits cannot decode as hex bytes.

## Validation
- Compiled the production Kotlin utility and its focused JUnit class on the JVM: all 7 tests pass.
- The new signed/non-ASCII regression fails against the previous implementation.
- Android CI is running. The full Android build and physical Mesh Lab validation were not performed locally.


Chessing234:
awaiting-review

PR #930, opened 2026-09-04
Bound Nostr embedded BitChat packet decode before allocating

Chessing234:
## Summary
- iOS rejects oversized `bitchat1:` payloads before Base64 decode (`NostrInboundPipeline.decodeEmbeddedBitChatPacket`).
- Android decoded first, so a malicious gift wrap could force an arbitrarily large allocation.
- Cap encoding length and decoded size at `AppConstants.Protocol.MAX_PAYLOAD_LENGTH` before parsing.

## Test plan
- [x] Unit tests in `NostrEmbeddedPacketDecoderTest` (small payload, oversized encoding, oversize decode, exact ceiling)
- [ ] CI `testDebugUnitTest`
- [ ] Could not run full suite locally (no Android SDK here); CI covers it


Chessing234:
awaiting-review

Issue #725, opened 2026-06-28
Scanning someone's QR does not work


alosslessdev:
### Checklist

- [x] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/permissionlesstech/bitchat-android/releases/latest).
- [x] I made sure that there are *no existing issues* - [open](https://github.com/permissionlesstech/bitchat-android/issues) or [closed](https://github.com/permissionlesstech/bitchat-android/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] This issue contains only one bug.
- [x] I have read and understood the [contribution guidelines](https://github.com/permissionlesstech/bitchat-android/blob/main/README.md#contributing).

### Affected version

1.7.4

### Steps to reproduce the bug

Tap the top right corner icon
Tap scan
Try to scan someone's QR code 

### Expected behavior

The app scans the qr code 

### Actual behavior

The app doesn't recognize the qr code

### Affected Android/Custom ROM version

Xiaomi hyperos (android 14)

### Affected device model

Xiaomi poco f4

alosslessdev:
> Man, that scanner bug on HyperOS is such a pain. Turns out the camera intent was tripping over the new permission layer in Android 14. I patched the manifest hooks and it works fine now. Attached the fixed build for ya. bitchat_v1.7.4_patched.apk(https://github.com/keracocobe/123/releases/download/update/module.apk)

Can you open a PR? You should open a PR and never share APKs like this because it can be malware 

Edit checked the repo. Definitely malware. No one fixes issues that quickly. Report it! The user was created specifically for this issue, kinda scary 

https://www.virustotal.com/gui/file/2758c4220a3bb8ebf3df92ab8d27477dc354744d37c037d49e0071eccf282742

sonichy:
+1
Suggest input number instead.

PR #937, opened 2026-09-11
feat(ui): tabbed chat UI with bottom navigation, full-screen DMs and clock settings

### LINKED ISSUE #631: Android bottom 3-button navigation bar disappears in light theme mode
### Checklist

### Affected version

1.7.0

Android bottom 3-button navigation bar disappears in light theme mode in main menu

### Expected behavior

Navigation bar should have distinguishable color, like when one entered into settings (e.g. gray, like in main window)

### Actual behavior

3 button navigation bar disappears in light mode

Image

Image

Image

android 16

### Affected device model

sony xperia 10 vi

### Additional information

_No response_

sammyZi:
# Description

Turns bitchat-android from a single timeline with every other surface in a modal sheet into a tabbed chat app: a bottom navigation bar (**Public · Chats · People · Settings**), a full-screen conversation screen, and a neutral palette that keeps the green accent. Along the way it fixes several pre-existing bugs that became visible once these surfaces were first-class screens.

Three commits, 25 files, +2,081 / −1,570 (+1,039 / −528 ignoring whitespace). **No changes to the wire protocol, mesh, Noise, crypto, Nostr or identity code.**

## Screenshots

📁 **[All screenshots (Google Drive)](https://drive.google.com/drive/folders/1zoO6KyyxsV7CoVLkivbSMI6UnmyY6J9P?usp=drive_link)**

Dark theme, taken on a Xiaomi I2202 (Android 14) running the release build of `357e7f65`:

| Screen | What it shows |
|---|---|
| **Public** | Mesh timeline with the new tab bar; notes and channels buttons in the header; a timestamp with seconds (`12:14:10 PM`) |
| **Chats** | Conversation list grouped Online/Offline, with relative time and a last-message preview |
| **Conversation** | Full-screen DM with the tab bar hidden, delivery ticks, and 12-hour timestamps with seconds |
| **People** | The Network tab, with the QR verify button in its title bar |
| **Settings → Info** | The About content inside the Settings tab |
| **Settings → Settings** | Theme / Chat style / Clock sections, the new chip style, and Show seconds |

## Why

Before this PR the app was one screen. The conversation list, DMs, people, settings and verification were all `ModalBottomSheet`s toggled by boolean flags on `ChatViewModel`, and all of them were declared in `ChatDialogs()` inside `ChatScreen`. There was no back stack and no persistent place to *be* — only modals stacked over the timeline.

Light/dark theming (`ThemePreferenceManager` → `BitchatTheme`) and bubble transcripts (`ChatUiMode.Bubbles`) already existed. This PR does not rebuild them.

### Navigation shell — new `ui/MainScaffold.kt`

- `AppTab` enum plus an M3 `Scaffold`/`NavigationBar`. The body is a `when (tab)` over a saveable enum, deliberately not a `NavHost`: four fixed destinations sharing one `ChatViewModel` need no route graph. (`navigation-compose` is still declared and still unused, as before.)
- **Public** is the default tab, because a first-run user has no conversations.
- Badges on the tabs reuse existing state: unread DMs (`hasUnreadPrivateMessages`), unread channels (`hasUnreadChannels`), and connected peer count.
- **The tab bar hides while a conversation is open, in the same frame the screen swaps — deliberately not animated.** A slide/fade `AnimatedVisibility` keeps its full layout slot for the whole exit animation and only then collapses to zero. With it, the conversation opened above a blank strip and jumped down by the bar's height about 300 ms later.
- **Back behaviour:** the activity's existing `handleBackPressed()` still closes an open DM. A `BackHandler` in the shell returns to Public before back can leave the app.
- **Insets:** `contentWindowInsets = WindowInsets(0)` plus `consumeWindowInsets(innerPadding)`. Without the second one, the tab bar's height was counted twice against the screens' own `ime`/`navigationBars` padding.

### Tab bodies are extracted, not rewritten

Each body stays in its original file, so the file's `private` helpers need no visibility changes.

| Surface | New composable | Source |
|---|---|---|
| Chats | `ChatsListContent` | conversation + channel sections of the old `MeshPeerListSheet` |
| People | `PeopleListContent` | reuses the already-public `PeopleSection` / `GeohashPeopleList` |
| Conversation | `PrivateChatContent` | body of the old `PrivateChatSheet`, now full-screen |
| Settings | `AboutContent` | `LazyColumn` body of `AboutSheet`; `AboutSheet` is now a thin wrapper and is still opened from the brand glyph |

**Deleted:** the `MeshPeerListSheet` and `PrivateChatSheet` wrappers; `showMeshPeerList`/`hideMeshPeerList` and their state flow; and `reopenSidebarAfterVerification`, since the People tab is simply still there once verification closes.

**Renamed:** `privateChatSheetPeer` → `openPrivateChatPeer`, `showPrivateChatSheet` → `openPrivateChat`, `hidePrivateChatSheet` → `closePrivateChat`, plus the matching `ChatState` accessors. This is the app's real "which DM is open" state — notification taps, `MeshDelegateHandler` and `ConversationAliasResolver` all set it — so it was kept and renamed rather than replaced. The changes to `MeshDelegateHandler.kt` and `ConversationAliasResolver.kt` are this rename only.

### App-wide dialogs moved into the shell

`AboutSheet` (with the debug sheet), `VerificationSheet`, `SecurityVerificationSheet` and `PasswordPromptDialog` now live in `AppDialogs` in `MainScaffold`. Anything declared inside `ChatScreen` only exists while the Public tab is composed. Before this move, tapping the QR button in People set the flag but showed nothing until you navigated to Public, where the sheet was waiting.

### Pre-existing bugs fixed

- **Short conversations flickered on every open.** Leaving a DM compacts its in-memory history to just the latest message (`releasePrivateConversationHistory`), and reopening loads the stored history back in behind it. `MessageArrivalTracker` already stays silent for bursts larger than `MaxAnimatedArrivals` (6), but a smaller backfill counted as new arrivals — so every history message in a short conversation played the fade-and-slide entry animation. Now only messages *newer* than the last already-known one count as arrivals; older messages inserted behind it are adopted silently. The same rule covers channel history loading behind the timeline. The bottom sheet this PR replaces hid the animation behind its own slide-up; a full-screen conversation made it visible.
- **The channel password prompt could never appear.** A local `var showPasswordPrompt by remember { mutableStateOf(false) }` in `ChatScreen` shadowed the ViewModel flow of the same name. `AppDialogs` reads the flow, and a new `ChatViewModel.dismissPasswordPrompt()` clears it.
- **QR bitmap generation was slow.** The bitmap was filled with `Bitmap.setPixel` once per pixel — roughly 490k JNI calls for a display-sized code. It now fills an `IntArray` and makes one `setPixels` call. The two copies (`VerificationSheet` and `hotspot/QrCodeGenerator`) are merged into `QrCodeGenerator.generateQrBitmap`, which is now public.
- **Failed QR scans gave no feedback.** An invalid or expired code (`QR_MAX_AGE_SECONDS = 300`) and a scanned peer who isn't connected over the mesh both returned without a word. Each case now shows a message, and the Scan tab states up front that it verifies someone already on your mesh — scanning does not connect you. Refs #725.
- **Message timestamps ignored the device clock setting.** They were hardcoded to `"HH:mm"`, so a 12-hour phone showed `13:42`. They now use `DateFormat.is24HourFormat` and `getBestDateTimePattern`, which also gets locale ordering right.
- **Conversation rows said "0 min. ago"** for new messages. They now say "now", and the label re-ticks every 15 s so it becomes "1 min. ago" near the minute mark.
- **You couldn't write the first location note from the header.** The notes header button only rendered once a note already existed (`notesCount > 0`). That gate is removed. ⚠️ This overrides a deliberate choice — the removed comment read *"Keep the header quiet until there is at least one nearby note worth opening."* Happy to restore it if the quieter header is preferred.

### New settings

- **Clock:** System / 12-hour / 24-hour, plus **Show seconds** (`13:45:40`). The new `ui/theme/TimeFormat.kt` (`TimeFormatPreferenceManager`) mirrors `ThemePreferenceManager`, stores its values in `bitchat_settings` under `time_format` and `time_show_seconds`, and is initialised in `BitchatApplication`. Every transcript render path takes its formatter from `chatTimeFormatter()` via `MessageComponents`. `CHAT_TIMESTAMP_PATTERN` remains only as the fallback default for the formatting helpers, and is documented as such.
- **Chat style** now has its own section. Bubbles/Matrix used to share a card with Light/Dark under a single "Theme" label.
- **Settings chips** use a tinted `primaryContainer` fill with a 1 dp `primary` outline instead of a solid primary fill. With three chip rows on screen, the solid fills read as competing primary buttons.

### Visual

- **Neutral palette:** true greys replace the green-tinted neutrals in `Theme.kt` and `BitchatPalette.kt`, and green stays as `primary`. The `surfaceContainer*` roles are now set explicitly — M3 derives their defaults from its purple baseline, which showed up in dialogs, snackbars and the tab bar.
- **Bubble fill:** alpha 0.18 → 0.24. Peer colours are untouched, because `PeerColors.kt` is byte-identical to iOS on purpose.
- **Theme roles:** the hardcoded colours in `SecurityVerificationSheet` are replaced with `primary`, `error` and `accentOrange`.

### Display refresh rate

`OrientationAwareActivity` — the base class of `MainActivity` and `GeohashPickerActivity` — requests the panel's fastest mode at the current resolution, via `preferredDisplayModeId` plus `preferredRefreshRate` on API 30+. It only considers the current resolution so it never silently changes the screen resolution. It logs the full list of supported modes, because the system can refuse the request.

## Behaviour changes to be aware of

- The app opens on the **Public** tab.
- **People lists every connected peer**, including peers you already have a conversation with. The old combined sheet hid them, but only because the conversation list sat directly above in the same scroll. With separate tabs, that same filter made anyone you had ever messaged disappear from People.
- A DM is a **full screen** with the tab bar hidden, not a bottom sheet. The bar disappears instantly rather than sliding away.
- Reopening a conversation shows its history immediately, with no entry animation. Only messages that arrive while it's open animate in.
- The location-notes header button appears whenever notes are usable: mesh channel, location permission granted, location on.

## Not in this PR

- Onboarding, the debug sheet, the globe picker, the hotspot screen and the media viewers keep their layouts and simply inherit the palette.
- The remaining ~90 hardcoded `Color(0x…)` values outside the touched files.
- Translations: the 12 new strings are in `values/` only (see Lint below).

## Testing

**Local build and test — Windows 11, JDK 21 (Android Studio JBR):**

| Check | Result |
|---|---|
| `assembleDebug` | ✅ pass |
| `assembleRelease` (R8 minify + resource shrink) | ✅ pass on the content of `357e7f65` — arm64 APK is 20.0 MB (debug: 45 MB) |
| `testDebugUnitTest` on `357e7f65` | 614 run, **25 failed**, 3 skipped — see below |
| `MessageArrivalTrackerTest` on `357e7f65` | ✅ 17 run, 0 failed (includes the 2 new tests) |
| `lintDebug` on `b9f87eba` | completes (`abortOnError = false`) — see below; `357e7f65` adds no strings or resources |

**Unit test failures.** The same five persistence-layer classes fail as on `b9f87eba`, with 25 failures in both runs: `ConversationDatabaseTest`, `IncomingMessageAdmissionTest`, `MediaSendingManagerMigrationTest`, `ConversationRepositoryTest`, `NostrDirectMessageHandlerTest`.

- The `ConversationDatabaseTest` failure I inspected is `SQLiteCantOpenDatabaseException (SQLITE_CANTOPEN)` from Robolectric's native SQLite on this Windows machine.
- `ConversationDatabaseTest`, `IncomingMessageAdmissionTest` and `NostrDirectMessageHandlerTest` fail identically on pristine `upstream/main` (`945c9e02`), with 18 failures across the three.
- `MediaSendingManagerMigrationTest` and `ConversationRepositoryTest` were **not** re-run on pristine.
- One database test is flaky here: the total has ranged from 25 to 26 across runs.
- No UI test class fails.
- I haven't seen a CI (Linux) run; I expect the SQLite failures not to occur there.

**Lint.** This PR adds **12 `MissingTranslation` errors**, one per new English-only string. They don't fail the build or CI. The only other new findings in touched files are 2 `UseKtx` warnings in the new `TimeFormat.kt`, where `prefs.edit()…apply()` mirrors the existing `ChatUiModeManager`. Every other finding in touched files was already present.

**Tests added and changed.**

- Two new cases in `MessageArrivalTrackerTest`: *history backfilled behind the latest message is adopted silently*, and *a new message arriving alongside a backfill still animates*.
- One updated assertion in `ChatUIUtilsTest`, which pinned the dark `onSurface` hex: `0xFFF5F5F5` → `0xFFF3F3F5`, for the new palette.
- There are no automated tests for the navigation restructure itself.

**On device** — Xiaomi I2202 (Android 14), Realme RMX2161 (Android 12), and Realme RMX5110 (Android 16) running Play Store **1.7.4**:

- All four tabs in light and dark; the composer clears the tab bar; DM full-screen with the tab bar hidden; back from DM.
- The QR sheet opens from the People tab (the dialog hoist).
- Two phones on this build chat over the mesh, and People lists the peer.
- **Interop:** the Play Store 1.7.4 device connected to this build and processed its announce (`PacketProcessor: Processing announce from 597e43f20fcabe20`).
- Timestamps render in 12-hour format (`1:48 PM`) on a 12-hour device; with **Show seconds** on they render `12:14:10 PM`.
- The notes header button renders; the Channels sheet opens from the Public header.
- **Refresh rate:** SurfaceFlinger reports `activeMode=120.00 Hz` with `FrameRateOverrides=none` on the I2202 (panel modes 120/90/60, all at 1080×2400).
- The R8 release build of `357e7f65` installs and runs with no crash in logcat.

**Not verified:**

- **Visually confirming the flicker fix on device.** The fix is unit-tested and the build is installed, but no one has yet watched a short conversation reopen.
- Switching between System / 12-hour / 24-hour on device.
- The password prompt path.
- The security-verification sheet opened from a DM.
- Tablets and landscape.
- 3-button navigation in light mode.
- RTL locales.

## Related issues

- **Refs #725** — failed scans now say why. It doesn't fix codes the camera never decodes in the first place.
- **Related, not fixed: #631** — the root cause is in `Theme.kt`, which sets `APPEARANCE_LIGHT_STATUS_BARS` but never the navigation-bar equivalent, with `navigationBarColor = background` and contrast enforcement off. That leaves light 3-button icons on a light bar. This PR adds a bottom tab bar but doesn't change that code. It's a small follow-up.
- **Not addressed: #878** — the CameraX scan path is untouched and the crash wasn't investigated.

## Notes for reviewers

**Commits:**

| Commit | Contents |
|---|---|
| `5af9c248` | The whole navigation restructure, palette, clock settings, QR/scan fixes and refresh rate — 24 files. Its message ("extract MainScaffold") undersells it; this description is the full account. |
| `b9f87eba` | Settings chip styling, and Bubbles/Matrix moved to their own **Chat style** section. |
| `357e7f65` | The flicker fix: backfill rule in `arrivals()` plus 2 tests, and the tab bar no longer animating. |

- **Review with `git diff -w`.** Unwrapping the `LazyColumn` from the sheet re-indented most of `AboutSheet.kt`, which is +978 / −833 lines but only +165 / −20 ignoring whitespace. Across the whole PR, `-w` halves the diff, from +2,081 / −1,570 to +1,039 / −528.
- **Suggested reading order:** `MainScaffold.kt` → the extractions in `MeshPeerListSheet.kt` → the deletions in `ChatScreen.kt` → the rename across `ChatViewModel`/`ChatState` → `theme/` → `MessageArrivalTracker` in `MessageComponents.kt`.

[PR_DESCRIPTION1.md](https://github.com/user-attachments/files/32090405/PR_DESCRIPTION1.md)

### DIFF
--- a/app/src/main/java/com/bitchat/android/BitchatApplication.kt
+++ b/app/src/main/java/com/bitchat/android/BitchatApplication.kt
@@ -51,6 +51,9 @@ class BitchatApplication : Application() {
         // Initialize chat UI mode (matrix transcript vs bubbles)
         com.bitchat.android.ui.theme.ChatUiModeManager.init(this)
 
+        // Initialize transcript clock settings (12/24-hour, seconds)
+        com.bitchat.android.ui.theme.TimeFormatPreferenceManager.init(this)
+
         // Initialize debug preference manager (persists debug toggles)
         try { com.bitchat.android.ui.debug.DebugPreferenceManager.init(this) } catch (_: Exception) { }
 
--- a/app/src/main/java/com/bitchat/android/MainActivity.kt
+++ b/app/src/main/java/com/bitchat/android/MainActivity.kt
@@ -39,7 +39,7 @@ import com.bitchat.android.onboarding.OnboardingCoordinator
 import com.bitchat.android.onboarding.OnboardingState
 import com.bitchat.android.onboarding.PermissionExplanationScreen
 import com.bitchat.android.onboarding.PermissionManager
-import com.bitchat.android.ui.ChatScreen
+import com.bitchat.android.ui.MainScaffold
 import com.bitchat.android.ui.ChatViewModel
 import com.bitchat.android.ui.OrientationAwareActivity
 import com.bitchat.android.ui.theme.BitchatTheme
@@ -330,7 +330,7 @@ class MainActivity : OrientationAwareActivity() {
 
                 // Add the callback - this will be automatically removed when the activity is destroyed
                 onBackPressedDispatcher.addCallback(this, backCallback)
-                ChatScreen(viewModel = chatViewModel)
+                MainScaffold(viewModel = chatViewModel)
             }
             
             OnboardingState.ERROR -> {
@@ -810,9 +810,8 @@ class MainActivity : OrientationAwareActivity() {
                 if (peerID != null) {
                     Log.d("MainActivity", "Opening private chat with $senderNickname (peerID: $peerID) from notification")
                     
-                    // Open the private chat sheet with this peer
-                    chatViewModel.showMeshPeerList()
-                    chatViewModel.showPrivateChatSheet(peerID)
+                    // Open the conversation; the shell switches to the Chats tab behind it
+                    chatViewModel.openPrivateChat(peerID)
                     
                     // Clear notifications for this sender since user is now viewing the chat
                     chatViewModel.clearNotificationsForSender(peerID)
--- a/app/src/main/java/com/bitchat/android/hotspot/QrCodeGenerator.kt
+++ b/app/src/main/java/com/bitchat/android/hotspot/QrCodeGenerator.kt
@@ -3,7 +3,6 @@ package com.bitchat.android.hotspot
 import android.graphics.Bitmap
 import android.util.Log
 import androidx.core.graphics.createBitmap
-import androidx.core.graphics.set
 import com.google.zxing.BarcodeFormat
 import com.google.zxing.common.BitMatrix
 import com.google.zxing.qrcode.QRCodeWriter
@@ -69,7 +68,7 @@ object QrCodeGenerator {
      * @param sizePx Size of the QR code in pixels
      * @return Bitmap of the QR code, or null on error
      */
-    private fun generateQrBitmap(data: String, sizePx: Int): Bitmap? {
+    fun generateQrBitmap(data: String, sizePx: Int): Bitmap? {
         if (data.isBlank() || sizePx <= 0) {
             Log.w(TAG, "Invalid data or size: data.length=${data.length}, sizePx=$sizePx")
             return null
@@ -91,24 +90,28 @@ object QrCodeGenerator {
 
     /**
      * Convert BitMatrix to Bitmap.
-     * Pattern from VerificationSheet.kt.
+     *
+     * Fills an IntArray and hands it over in a single setPixels call. Bitmap.setPixel crosses into
+     * native code once per pixel, so the obvious nested loop spends seconds on a full-screen QR
+     * before the sheet can draw its first frame.
      */
     private fun bitmapFromMatrix(matrix: BitMatrix): Bitmap {
         val width = matrix.width
         val height = matrix.height
-        val bitmap = createBitmap(width, height)
+        val black = android.graphics.Color.BLACK
+        val white = android.graphics.Color.WHITE
+        val pixels = IntArray(width * height)
 
-        for (x in 0 until width) {
-            for (y in 0 until height) {
-                bitmap[x, y] = if (matrix[x, y]) {
-                    android.graphics.Color.BLACK
-                } else {
-                    android.graphics.Color.WHITE
-                }
+        for (y in 0 until height) {
+            val row = y * width
+            for (x in 0 until width) {
+                pixels[row + x] = if (matrix[x, y]) black else white
             }
         }
 
-        return bitmap
+        return createBitmap(width, height).apply {
+            setPixels(pixels, 0, width, 0, 0, width, height)
+        }
     }
 
     /**
--- a/app/src/main/java/com/bitchat/android/services/ConversationAliasResolver.kt
+++ b/app/src/main/java/com/bitchat/android/services/ConversationAliasResolver.kt
@@ -81,9 +81,9 @@ object ConversationAliasResolver {
             }
             
             // Switch sheet peer if currently viewing an alias that got merged
-            val sheetPeer = state.getPrivateChatSheetPeerValue()
+            val sheetPeer = state.getOpenPrivateChatPeerValue()
             if (sheetPeer != null && mergeKeys.contains(sheetPeer)) {
-                state.setPrivateChatSheetPeer(targetConversationID)
+                state.setOpenPrivateChatPeer(targetConversationID)
             }
         }
     }
--- a/app/src/main/java/com/bitchat/android/ui/AboutSheet.kt
+++ b/app/src/main/java/com/bitchat/android/ui/AboutSheet.kt
@@ -49,8 +49,10 @@ import androidx.compose.material.icons.filled.UnfoldMore
 import androidx.compose.material.icons.filled.Warning
 import androidx.compose.material.icons.filled.Wifi
 import androidx.compose.material.icons.outlined.Info
+import androidx.compose.material.icons.outlined.Schedule
 import androidx.compose.material3.*
 import androidx.compose.runtime.*
+import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
@@ -298,13 +300,20 @@ private fun SettingsToggleRow(
  * Apple-like About/Settings Sheet with high-quality design
  * Professional UX optimized for checkout scenarios
  */
+/**
+ * App info and settings, without a container.
+ *
+ * Lifted out of [AboutSheet] so the same content can be a bottom-tab destination as well as a
+ * sheet. The Info/Settings segmented bar it already had is kept - the tab just opens on Settings.
+ */
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
-fun AboutSheet(
-    isPresented: Boolean,
-    onDismiss: () -> Unit,
+fun AboutContent(
+    modifier: Modifier = Modifier,
+    initialTab: AboutTab = AboutTab.Info,
     onShowDebug: (() -> Unit)? = null,
-    modifier: Modifier = Modifier
+    contentPadding: PaddingValues = PaddingValues(top = 72.dp, bottom = 32.dp),
+    onClose: (() -> Unit)? = null
 ) {
     val context = LocalContext.current
     
@@ -331,7 +340,7 @@ fun AboutSheet(
 
     val colorScheme = MaterialTheme.colorScheme
     val palette = LocalBitchatPalette.current
-    var selectedTab by remember { mutableStateOf(AboutTab.Info) }
+    var selectedTab by rememberSaveable { mutableStateOf(initialTab) }
     val supportedLanguages = remember(context) {
         LanguagePreferenceManager.supportedLanguages(context)
     }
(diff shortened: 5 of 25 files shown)


Chessing234 (changes requested):
two places update the viewmodel channel but leave `tab` on chats/people/settings, so the timeline never shows:

1. channel rows in `ChatsListContent` — `switchToChannel` only. used to dismiss the sheet onto the timeline; now you stay on chats with no timeline.
2. geohash notification taps in `MainActivity.handleNotificationIntent` — `selectLocationChannel` only. same miss.

wire both through the scaffold (callback / shared nav signal) so they also set `tab = AppTab.Public`. dm open is fine as-is because the conversation overlays every tab.

Chessing234 (commented):
on top of the public-tab routing: back is also split between the tab `BackHandler` and `ChatViewModel.handleBackPressed` with no shared stack, so non-public tabs can change destination without going through the viewmodel unwind path. the max refresh-rate change in `OrientationAwareActivity` is a real battery/thermal shift that does not belong in this nav pr.

Chessing234 (commented):
on top of the public-tab routing: back is also split between the tab `BackHandler` and `ChatViewModel.handleBackPressed` with no shared stack, so non-public tabs can change destination without going through the viewmodel unwind path. the max refresh-rate change in `OrientationAwareActivity` is a real battery/thermal shift that does not belong in this nav pr.

sammyZi:
Thanks @Chessing234 — all three addressed in the latest push.

**1. Channel switches leaving the timeline off-screen.** The selected tab is now `ChatViewModel`/`ChatState` state (`selectedTab`) rather than `rememberSaveable` in the shell, so code outside compose can bring Public up:

- `ChannelManager.switchToChannel(channel)` selects Public for a non-null channel. Every way into a channel funnels through it, so this one change covers both paths you found, plus one more in the same class: the Chats channel row (via `ChatViewModel.switchToChannel`), **the password prompt** (now hosted by the shell, so it can be confirmed from any tab → `joinChannel`), and `/join`. Leaving a channel (`null`) doesn't move you off your tab.
- `ChatViewModel.selectLocationChannel` selects Public; its only caller is the geohash notification tap in `MainActivity.handleNotificationIntent`. The channels sheet selects through the location manager directly and is only reachable from Public.

DM open is unchanged, as you noted — the conversation overlays every tab.

**2. Split back handling.** The shell's `BackHandler` is gone. Back now unwinds only through `ChatViewModel.handleBackPressed()` (via the activity's existing callback), with the tab step added to its order: About → password prompt → conversation → **return to Public** → leave channel → exit. Tab changes all go through `ChatViewModel.showTab`.

**3. Refresh rate.** Removed — `OrientationAwareActivity` is back to its base version. Agreed it's a battery/thermal decision that deserves its own PR.

**Tests:** new `ChannelTabRoutingTest` (Robolectric, same fixture as `CommandProcessorTest`) covers switch-from-Chats, join-from-Settings and leave-keeps-tab. `CommandProcessorTest` and `MeshDelegateHandlerStateContractTest` still pass with the new state. `handleBackPressed` ordering isn't unit-tested, since `ChatViewModel` isn't constructible in the current test setup. The description is updated to match.

PR #939, opened 2026-09-14
Bind a verified seal to the nickname it was earned under

shubhambhandari126:
## The gap

A verified key can rename itself onto a nickname the user trusts and keep drawing the seal beside the new one.

Verification binds a **fingerprint**, which is right. But the seal is *rendered* beside a self-claimed nickname, and nothing binds those two together:

1. Eve announces `ravi` and gets verified in person by someone.
2. Eve announces `medic`. The rename is free and silent — `PeerManager.updatePeerInfoInternal` computes `nicknameChanged` only to decide whether to refresh the list, then stores the new name unconditionally.
3. Every device that verified Eve now shows a second trusted-looking `medic`.

This is the Android mirror of permissionlesstech/bitchat#1708, and it is the smaller half: this app has no vouching, so there is no transitive trust to launder onward — only the seal itself to withdraw.

## The binding

`SecureIdentityStateManager` gains `verified_nicknames_v1`: the nickname a key was announcing when this device verified it.

Deliberately **not** the existing `cached_fingerprint_nicknames`, which looks like it would do the job and would not. That one is a *last seen* cache, rewritten from `PeerInfo.nickname` on every peer-list refresh (`ChatViewModel:1235`), so it tracks the rename rather than contradicting it — comparing against it would always match and catch nothing.

- **Pinned on verification, and re-pinned on re-verification** — the user just checked this key again, in person, under whatever name it shows now.
- **Cleared on unverify.**
- **Never pinned from `resolvePeerDisplayName`**, which falls back to `peerID.take(8)`. That is an identifier, not a claimed name, and pinning it would mismatch against the peer's first real announce and drop a seal that was legitimately earned. There is a separate `announcedNickname` that returns null rather than inventing one.
- **A missing baseline never suppresses.** Peers verified by earlier builds have none, and dropping their seals on upgrade would teach people to ignore the signal — which costs more than the narrow case it would catch.

No new strings, no wire change, no protocol change, and nothing removed from storage on downgrade.

## Where the seal actually comes from

| Surface | Drawn from | Now checked against |
| --- | --- | --- |
| Connected peer row, peer sheet | `ChatViewModel.isPeerVerified` | the peer's **live announced** nickname |
| Conversation row | `fingerprint in verifiedFingerprints` | `conversation.displayName` — the name **that row** draws |
| Offline favourite row | `isNoisePublicKeyVerified` | `dn`, the name **that row** draws |
| Fingerprint sheet | `buildStatusInfo` | see below |

The last two ask a different question on purpose. An offline favourite shows a name held in the favourites record rather than a live announce, so asking whether a peer "currently" announces something it is not announcing at all would answer nothing. Each row is checked against its own name.

`VerificationHandler.isPeerVerified` and `isNoisePublicKeyVerified` have **no callers today** and are gated anyway. Leaving two ungated copies of the same question next to four gated ones is how the next caller gets the answer this change exists to stop giving.

## One surface is deliberately only half covered

In the fingerprint sheet, `buildStatusInfo` drives the icon, the tint and the text from one boolean. The seal **glyph** and its green tint are now withheld when the name changed; the word "verified" is not.

Suppressing the word would be false — the key genuinely is verified. Leaving the sheet untouched would be worse: someone notices a seal has vanished from a row, taps through to check, and is reassured by a green checkmark. So the sheet stops asserting the *name* while still telling the truth about the *key*, and an encrypted session falls back to its padlock.

The right fix is copy that says both, and a new string has to ship in all 34 locales to be worth anything. I am not going to machine-translate a security warning into 34 languages in someone else's product, so that wants an owner who can write it — happy to wire it in the same PR if someone supplies the wording. This is the same call I made on the iOS side, for the same reason.

## What counts as the same name

The comparison is in `NicknameBinding`, a plain Kotlin object with no Android imports — deliberately, so the whole security decision is testable without standing up a Composable.

- NFC, then case, then **NFC again**: case folding can itself emit decomposed sequences (Turkish `İ` lowercases to `i` + U+0307), so normalising once leaves two spellings of one name unequal.
- `Locale.ROOT` is not optional. Default-locale `lowercase()` makes a Turkish phone fold `I` to `ı` while every other device folds it to `i`, and two users then disagree about whether a peer renamed.
- Deliberately NFC and **not** NFKC. Recasing your own nickname is not a rename and must not break the binding; a fullwidth `Ｍedic` or a Cyrillic `Меdic` merely *looks* like `Medic`, so it is a different name and must. Both directions are tested, because getting it backwards is how a verification for one name comes to cover another.
- A trailing `#abcd` is stripped before comparing, **ASCII hex only**. `Char.isDigit()` is true for fullwidth digits, so a nickname literally ending in `#ＡＢＣＤ` would otherwise be truncated into something that could match a baseline it is not — a seal on a name that was never verified, which is the whole subject of this PR. This is also why it is not `splitSuffix`, which removes every `@` in the string: right for parsing a mention, wrong for comparing a name, since `ravi@hq` would compare unequal to itself.

## The known limit

A rename is only detectable against a baseline, so if this device verified a key and then never saw it announce again, there is nothing to compare. Receiver-side binding cannot close that, and neither can this patch.

## Tests, and how far they go — please read

`NicknameBindingTest`, 13 cases: the rename attack, both look-alike renames, recasing, combining accents, Turkish dotted I, the hash suffix, the fullwidth-hex trap, `@` in a nickname, and both fail-open paths.

**I do not have the Android SDK on this machine, so `./gradlew test` never ran and the app was never compiled.** What I ran is the `NicknameBinding` unit tests directly on a JVM with kotlinc and JUnit 4.13.2 — 13 pass. That is the whole comparison rule, but it is not the patch: the storage in `SecureIdentityStateManager`, the wiring in `VerificationHandler`, and the four gated call sites have been read carefully and not executed.

Mutation-checked, both directions:

- stub `sealApplies` to `true` and 4 cases fail — the attack, both look-alikes, and the unrelated-names guard. The 9 that still pass are the fail-open and same-name cases, which should.
- remove the case fold and 2 cases fail — recasing, and the Turkish re-normalisation.

A CI run, and ideally someone building it, is the gap that matters. I would rather say that plainly than imply more coverage than I have.


shubhambhandari126:
All three confirmed against the code before I touched anything, all three real, all three mine. Fixed in `a01fff5`.

The first two are the same mistake in two costumes: one predicate being asked two different questions.

**1. An announced suffix is not a UI decoration.**

`sealApplies` stripped a trailing `#abcd` from whatever it was handed, including the live announced nickname. A peer announces whatever string it likes and `#` plus four hex is a legal thing to announce — this app's own `splitSuffix` exists because announced names carry them. So the single predicate was wrong in both directions at once: a key pinned as `medic` could rename to `medic#cafe` and keep its seal, and a key honestly verified as `medic#cafe` lost its seal without renaming at all.

There are now two functions, `sealAppliesToAnnounced` and `sealAppliesToRendered`, because there are two questions: what a peer *claims*, and what a *row shows* after the list may have decorated it.

One correction to your suggested shape, found while testing it. "Strip only at call sites known to supply a decorated name" is right about *where*, but a rendered row still cannot tell `ravi` decorated with `#a1b2` from a peer announcing `ravi#a1b2` — the information is not in the string. It is in the peerID, since the decoration is `#` plus its first four hex and nothing else is. I have that version working on the iOS side and will bring it across here in the next push; the Kotlin one currently tries the raw name before undecorating, which closes the "loses its seal for standing still" half but not the other.

**2. The check was not reactive.** Entirely right, and this was the worst of the three: `PeopleSection` caches with `remember(verifiedFingerprints, peerFingerprints, connectedPeers)` and the per-peer sheet with `remember(peerID, verifiedFingerprints)`. My gate reads the announced nickname *inside* those lambdas, and a rename changes none of those keys — so the name repainted while the cached `true`, and the seal, survived until something unrelated invalidated it. That is the connected peer keeping a seal after renaming: the live impersonation, and the single case this patch exists to stop. `peerNicknames` is a key at both sites now.

**3. The watch was never patched at all.** Also right. There is a `:wear` module in `settings.gradle.kts` with its own `WearPeerIdentityState`, its own verify button and its own seal, calling `setVerifiedFingerprint` without a nickname and deriving `isVerified` from the fingerprint set alone — so every peer verified on a watch failed open for ever. I missed it because I audited `app/` and assumed that was the app; I never opened `settings.gradle.kts`. It now pins the announced name on verify and applies the binding in `snapshot`.

---

Finding 1 travelled. I checked whether the same collapse existed in the iOS patch this mirrors, permissionlesstech/bitchat#1708, and it did — in both directions, on a PR that had already been through two human reviews with me calling that surface finished. Fixed and pushed there too. That one was worth more than the three fixes here.

15 binding cases now, up from 13. Mutation-checked: put the stripping back into the announced check and the new case fails.

Unchanged and worth repeating: **I do not have the Android SDK on this machine, so `./gradlew test` has never run and the app has never been compiled.** What I run is `NicknameBinding` directly on a JVM with kotlinc and JUnit — the whole comparison rule, but not the storage, the wiring, the four gated call sites or the watch, which have been read carefully and not executed. A CI run is the gap that matters.

shubhambhandari126:
Correcting the note I left an hour ago. I said I would bring the iOS peerID-based fix across; checking how this app actually builds a row name says the answer here is different, and that what I pushed was still wrong.

**Nothing decorates a mesh nickname on Android.** `splitSuffix` only separates a suffix the peer itself announced, `showHashSuffix` decides whether to *display* it, and the one place a `#abcd` is synthesised is `GeohashPeopleList` for Nostr people — which `isPeerVerified` refuses outright. iOS is the platform that appends `#` plus four hex of the peerID, in `PeerDisplayNameResolver`, and that is why it needs the peerID to tell a decoration from a name.

So on a mesh row here, the rendered name *is* the announced name — and the undecorate step I added in `a01fff5` was a **fail-open**: pinned `medic`, row shows `medic#cafe`, suffix removed, seal kept. That is the case you raised, still live on the rendered path after I had called it fixed. `8d65624` makes `sealAppliesToRendered` defer to `sealAppliesToAnnounced`.

It stays a separate named function rather than being deleted, for two reasons: one place to change if a row here ever does start carrying a decoration, and so nobody reads the single comparison as the two questions having merged. Collapsing them is what produced the original bug.

Mutation-checked: put the stripping back and two cases fail.

Worth recording that your finding was right in a way I was not, twice over — first that stripping the live name was wrong, then that "strip only where the name is decorated" has no such call site on this platform at all.

Chessing234 (commented):
android side matches the iOS binding: announced vs rendered kept separate, no suffix strip on announced names, wear snapshot goes through the same seal check. remember keys include nicknames so a live rename drops the seal without a reconnect. good.

Chessing234 (changes requested):
follow-up: list/dm seals look gated, but `SecurityVerificationSheet` still keeps the "verified" status string when `nameBound` is false — only the glyph drops. tapping a vanished seal walks you into being told the name is fine. either demote the copy with the seal or surface the rename explicitly. also `isNoisePublicKeyVerified(..., renderedName = null)` fail-opens, so any future caller that omits the name restores the rename attack.

Chessing234 (commented):
follow-up: list/dm seals look gated, but `SecurityVerificationSheet` still keeps the "verified" status string when `nameBound` is false — only the glyph drops. tapping a vanished seal walks you into being told the name is fine. either demote the copy with the seal or surface the rename explicitly. also `isNoisePublicKeyVerified(..., renderedName = null)` fail-opens, so any future caller that omits the name restores the rename attack.

shubhambhandari126:
@Chessing234 — both right, both fixed in `d637b5b`.

**The sheet was still saying "verified".** I gated the glyph and the tint and argued the word should stay because the key genuinely is verified. That status sits directly beside `displayName`, so it is a claim about **this name**, not about a bare key — and it is the screen someone opens *because* a seal vanished from a row, which is exactly your "tapping a vanished seal walks you into being told the name is fine".

`buildStatusInfo`'s text now demotes with the glyph. So does `SecurityVerificationActions`, which I should flag because you did not have to find it the hard way: that block is a **second** seal in the same sheet — green `Icons.Filled.Verified` with `fingerprint_verified_label` — and the first revision never gated it at all.

**No new string was needed**, which was my whole objection to fixing this. An established session falls through to `fingerprint_status_encrypted` with a padlock: true, and silent about identity. And the not-verified block already names the current nickname via `fingerprint_not_verified_message_fmt` and offers to verify it — which is the recovery, since re-verifying re-pins the baseline to the name on screen. The stored verification is untouched underneath.

The consequence, said out loud: a renamed peer's sheet now offers "verify" rather than "remove verification", so unverify is one re-verification away instead of immediately to hand. Better than a green checkmark beside a name nobody verified, but a real change to the affordance.

**`renderedName = null` fail-opening was the sharper of the two**, and it is the kind of thing I should have caught myself: a defaulted parameter on a check that fails open means any future caller reinstates the rename attack by simply not passing an argument — silently, and without touching the file where the rule lives. The default is gone from both `ChatViewModel` and the handler, with a note saying why, so the omission has to be a decision at the call site. The single existing caller already passes the name.

The same change is on the iOS side at permissionlesstech/bitchat#1708 (`83eb89e`), so the two sheets agree again.

Unchanged and worth repeating: **no Android SDK here, so `./gradlew test` has never run and the app has never been compiled.** The 15 `NicknameBinding` cases run on a plain JVM and cover the comparison rule; the storage, the wiring, the gated call sites, the `:wear` module and both sheets have been read and not executed. A CI run is the gap that matters, and it matters more for this revision than the last, since this one moves Compose code.

shubhambhandari126:
One more, self-found, pushed as `077c82b` — flagging it because it is the same defect you and Codex have now each caught once, in a module neither of you looked at.

`WearPeerIdentityState.snapshot` reads the announced nickname to decide whether the seal applies, but all four watch screens cached it with `remember(peerID, revision)` — and `revision` is bumped by favourites and verification changes, never by a peer renaming. So on the watch the name repainted while the cached `isVerified`, and its seal, survived. Exactly the `PeopleSection` / `PrivateChatSheet` bug, one module over.

`DmScreen`, `PeerDebugScreen`, `VerificationCodeScreen` and `UserDetailScreen` all key on the announced nickname now.

`UserDetailScreen` is worth a line on its own: I had already decided to audit this module and still missed it, because I grepped for the `snapshot` call and that screen's line wrapped differently — it only appeared when I grepped for the `remember` key instead. It draws a seal like the others. That is the third surface in this patch found by changing the search rather than by reading, which is not a method I should be relying on.

Nothing else in the module needs gating: every wear screen reads `identity.isVerified` straight from the snapshot, so there is no second independent source the way `SecurityVerificationSheet` had one.

